PR #1, opened 2015-07-31
Migration to Tomcat 8 / Java 8.

camarasu:
Attention, Java source and target change from 1.7 to 1.8.
The new bcprov-jdk15on dependency may have to be manually installed. It can be downloaded at this address: https://www.bouncycastle.org/download/bcprov-jdk15on-152.jar

### DIFF
--- a/Portal/pom.xml
+++ b/Portal/pom.xml
@@ -37,30 +37,30 @@ knowledge of the CeCILL-B license and that you accept its terms.
     <groupId>fr.insalyon.creatis</groupId>
     <artifactId>vip</artifactId>
     <packaging>pom</packaging>
-    <version>1.14</version>
+    <version>1.15</version>
     <name>VIP</name>
     
     <properties>
-        <vip.core.version>1.14</vip.core.version>
-        <vip.portal.version>1.14</vip.portal.version>
-        <vip.docs.version>1.6</vip.docs.version>
-        <vip.application.version>1.12</vip.application.version>
-        <vip.datamanager.version>1.12</vip.datamanager.version>
-        <vip.datamanager-applet.version>1.13</vip.datamanager-applet.version>
-        <vip.gatelab.version>1.9</vip.gatelab.version>
-        <vip.gatelab-applet.version>1.13</vip.gatelab-applet.version>
+        <vip.core.version>1.15</vip.core.version>
+        <vip.portal.version>1.15</vip.portal.version>
+        <vip.docs.version>1.15</vip.docs.version>
+        <vip.application.version>1.15</vip.application.version>
+        <vip.datamanager.version>1.15</vip.datamanager.version>
+        <vip.datamanager-applet.version>1.15</vip.datamanager-applet.version>
+        <vip.gatelab.version>1.15</vip.gatelab.version>
+        <vip.gatelab-applet.version>1.15</vip.gatelab-applet.version>
         <vip.simulation-gui.version>0.4.1</vip.simulation-gui.version>
         <vip.models.version>0.9.1</vip.models.version>
         <vip.simulated-data.version>1.3.1</vip.simulated-data.version>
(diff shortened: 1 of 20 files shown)


glatard:
Can you test a complete workflow in a Tomcat 8 deployment, including:
* proxy creation.
* data transfers to/from the portal
* application launch and monitoring
* Dropbox links
?

Portal/pom.xml
```diff
@@ -152,8 +152,8 @@ knowledge of the CeCILL-B license and that you accept its terms.
                 <artifactId>maven-compiler-plugin</artifactId>
                 <version>3.0</version>
                 <configuration>
-                    <source>1.7</source>
-                    <target>1.7</target>
+                    <source>1.8</source>
```
glatard: Source and target version should be changed in project modules too.

Portal/vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/business/proxy/ProxyClient.java
```diff
@@ -353,12 +354,19 @@ private void printKey(PrivateKey key, PrintStream out)
         out.println("-----BEGIN RSA PRIVATE KEY-----");
         ByteArrayInputStream inStream = new ByteArrayInputStream(key.getEncoded());
         ASN1InputStream derInputStream = new ASN1InputStream(inStream);
-        DERObject keyInfo = derInputStream.readObject();
+        ASN1Primitive keyInfo = derInputStream.readObject();
+        /*
```
glatard: Remove useless comment?

Portal/pom.xml
camarasu: Done

camarasu:
Tests for proxy creation, data transfers to/from the portal and application launch/monitoring are OK.

Dropbox link fails both with the new and the deplyed version of the portal. Therefore, I opened a specific ticket for it (Maven Issue #2687)

Portal/vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/business/proxy/ProxyClient.java
camarasu: Done.

camarasu:
I keep this branch named "develop" for the current pull-request. I'll create a new "feature" branch for new commits.

camarasu:
I changed the version number of the portal (and modified modules accordingly), but, on a second thought, we may only want to change them when the next release (1.15) is ready...

Portal/pom.xml
```diff
@@ -37,30 +37,30 @@ knowledge of the CeCILL-B license and that you accept its terms.
...
-        <vip.portal.version>1.14</vip.portal.version>
-        <vip.docs.version>1.6</vip.docs.version>
-        <vip.application.version>1.12</vip.application.version>
-        <vip.datamanager.version>1.12</vip.datamanager.version>
-        <vip.datamanager-applet.version>1.13</vip.datamanager-applet.version>
-        <vip.gatelab.version>1.9</vip.gatelab.version>
-        <vip.gatelab-applet.version>1.13</vip.gatelab-applet.version>
+        <vip.core.version>1.15</vip.core.version>
+        <vip.portal.version>1.15</vip.portal.version>
+        <vip.docs.version>1.15</vip.docs.version>
```
glatard: Only the version of modules that were updated in the current version must be changed. For instance, if vip-docs was not updated in 1.15, it must keep its previous (1.6) version.

camarasu: Actually, I changed the version of all modules that have been re-compiled with Java 8 because their pom files have changed (as well as the corresponding jars, of course). If you believe these changes should not impact version numbers, I'll set them back to their previous version.

PR #2, opened 2015-09-07
Gatetest

camarasu:
Redmine issue #2478

### DIFF
--- a/Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/client/ApplicationModule.java
+++ b/Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/client/ApplicationModule.java
@@ -96,8 +96,8 @@ public class ApplicationModule extends Module {
 
     @Override
     public void load() {
-        //reservedClasses = new ArrayList<> ();
-        reservedClasses=reservedGateClasses;
+        reservedClasses = new ArrayList<String>();
+        reservedClasses.addAll(reservedGateClasses);
         reservedClasses.addAll(reservedTestClasses);
         CoreModule.addGeneralApplicationParser(new ApplicationHomeParser());
         CoreModule.addSystemApplicationParser(new ApplicationSystemParser());
--- a/Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/client/ApplicationModule.java
+++ b/Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/client/ApplicationModule.java
@@ -60,6 +60,8 @@ import java.util.List;
 public class ApplicationModule extends Module {
 
     public static List<String> reservedClasses;
+    public static List<String> reservedGateClasses;
+    public static List<String> reservedTestClasses;
 
      public ApplicationModule() {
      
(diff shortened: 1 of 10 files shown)


Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/client/ApplicationModule.java
```diff
@@ -72,17 +74,31 @@ public void onFailure(Throwable caught) {
...
+            public void onFailure(Throwable caught) {
+                Layout.getInstance().setWarningMessage("Unable to load applet gatelab classes:<br />" + caught.getMessage());
+            }
+
+            @Override
+            public void onSuccess(List<String> result) {
+             reservedTestClasses=result;
+            }
+        };
+        ApplicationService.Util.getInstance().getAppletGateLabTestClasses(callbackTest);
```
glatard: Do you really need 2 remote calls?

Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/client/ApplicationModule.java
```diff
@@ -72,17 +74,31 @@ public void onFailure(Throwable caught) {
...
+          
     }
 
     @Override
     public void load() {
-
+        reservedClasses = new ArrayList<String>();
+        reservedClasses.addAll(reservedGateClasses);
```
glatard: At this point there is no guarantee that reservedGateClasses was initialized: the call to getApletGateLab* done in the constructor could still be in progress. This is a bug, but it was already present before your modifications.

Portal/vip-gatelab/src/main/java/fr/insalyon/creatis/vip/gatelab/client/GateLabModule.java
```diff
@@ -47,18 +47,24 @@
...
         if (CoreModule.user.isSystemAdministrator()
                 || CoreModule.user.hasGroupAccess(GateLabConstants.GROUP_GATELAB)) {
-            for(String GateLabClass:ApplicationModule.reservedClasses){
+            for(String GateLabClass:ApplicationModule.reservedGateClasses){
             CoreModule.addApplicationsTileGrid(new GateLabTileGrid(GateLabClass));
             }
         }
+        if (CoreModule.user.isSystemAdministrator()
+                || CoreModule.user.hasGroupAccess(GateLabConstants.GROUP_GATELABTEST)) {
+            for(String GateLabClass:ApplicationModule.reservedTestClasses){
```
glatard: This solution works for 2 GATE classes, but it is very ad-hoc. I would rather do the following:
for(String GateLabClass:ApplicationModule.reservedGateClasses){
  // Check which groups can access this class.
  // Check if user has access to one of these groups. If yes, addApplicationsTileGrid. 
}

Portal/vip-gatelab/src/main/java/fr/insalyon/creatis/vip/gatelab/client/GateLabModule.java
```diff
@@ -47,18 +47,24 @@
...
+                CoreModule.addApplicationsTileGrid(new GateLabTileGrid(GateLabClass));
+            }
+        }
```
glatard: As a general comment, I think a cleaner way to implement this would be:
1. to add a "gateClasses" boolean to ApplicationService.Util.getInstance().getClasses to retrieve either the general classes or the gate classes. Only the classes that the user can access are returned (as in the existing implementation).
2. to implement ApplicationModule.load and GateLabModule.load in the same way, calling ApplicationService.Util.getInstance().getClasses with "true" or "false" for "gateClasses".

That would get rid of the reservedClass buggy code in the ApplicationModule constructor (see comment above). It would also do security checks server side rather than client side.

Portal/vip-gatelab/src/main/java/fr/insalyon/creatis/vip/gatelab/client/GateLabModule.java
```diff
@@ -47,18 +47,24 @@
...
     public GateLabModule() {
-      new ApplicationModule();
+      //new ApplicationModule();
     }
 
     @Override
     public void load() {
-
+        
         if (CoreModule.user.isSystemAdministrator()
```
glatard: Such tests should be done server-side, otherwise malicious clients can pretend to be admins (but this was also present before your modification).

Portal/vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/business/Server.java
```diff
@@ -214,7 +218,19 @@ private Server() {
...
             appletGateLabClasses = config.getList(CoreConstants.APPLET_GATELAB_CLASSES, appletGateLabCl);
-
+            
+            List<String> appletGateLabTestCl = new ArrayList<String>();
+            appletGateLabTestCl.add("GateLab Test");
+            appletGateLabTestClasses = config.getList(CoreConstants.APPLET_GATELABTEST_CLASSES, appletGateLabTestCl);
+           
+            reservedClasses = new HashMap<String, Integer>();
+            for (final String gateClass : appletGateLabClasses) {
+                this.reservedClasses.put(gateClass, 0);
```
glatard: Using an enum type instead of 0/1 would be easier to read.

PR #12, opened 2015-11-12
Redmine #2678: support for Carmin v0.1.

glatard:
Carmin is an API defined in FLI-IAM to access processing platforms. It is implemented as SOAP web-services following the specification in https://github.com/fli-iam/api


Portal/vip-api/src/main/java/fr/insalyon/creatis/vip/api/business/ExecutionBusiness.java
```diff
@@ -31,25 +31,329 @@
...
+
+            TransferPoolBusiness tpb = new TransferPoolBusiness();
+            WorkflowBusiness wb = new WorkflowBusiness();
+            List<InOutData> outputs = wb.getOutputData(executionId, getUser().getFolder());
+            for (InOutData output : outputs) {
+                
+                String operationId = tpb.downloadFile(getUser(), output.getPath());
+                
+                String url = getRequest().getRequestURL() + "/../fr.insalyon.creatis.vip.portal.Main/filedownloadservice?operationid=" + operationId; 
+                URL u = new URL(url); // just to check that it is a well-formed URL
```
camarasu: You don't use the newly created URL u in the following. Is this normal ?

glatard: Yes, because the method returns a string. The URL is built just to check that it is well formed. I could also return u.toString(), but this is equivalent.

camarasu:
Great job, really !

The only problem I detected on the portal is that the persona sign-in is no longer available and, probably related, there is a "Script load failed: /js/login-persona.js" message when logging out using Firefox.

Except for that, I also have a silly question: besides checking the /api /api?wsdl pages, how do you currently test the API itself ?

glatard:
This script load error is not coming from the API. It is specific to deployments in Tomcat servers managed by users. It should work in production.

A test suite is available at https://github.com/glatard/api/tree/master/tests (should be merged soon to https://github.com/fli-iam/api/tree/master/tests).

camarasu:
OK, great !
I think the pull request can be merged if that's ok with you.

PR #14, opened 2015-11-20
bcprov exclusions

camarasu:
BCprov exclusions and AntiResourceLocking


camarasu:
OK: removed attribute and changed path (it does not need to correspond to the version of the portal).

PR #20, opened 2016-02-05
Jar sign

camarasu:
Removed old keystore files (keystore files should not be public).


camarasu:
Sorry, I've just noticed I have some extra usless commits. I'll close the pull request, clean up and do another one.

PR #21, opened 2016-02-16
Moteur Load-Balancing

camarasu:
The Load Balancing is done based on the number of workflows a given engine is already running. 
There is a 'hack' that needs to be removed before going into production.


Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/client/view/system/classes/EditClassLayout.java
```diff
@@ -131,14 +133,14 @@ public void execute(Boolean value) {
...
+     * @param engines Class engines
      */
-    public void setClass(String name, String groups, String engine) {
+    public void setClass(String name, String groups, String engines) {
 
         if (name != null) {
             this.nameField.setValue(name);
             this.nameField.setDisabled(true);
-            this.enginesPickList.setValue(engine);
+            this.enginesPickList.setValues(engines.split(", "));
```
glatard: You could split by "," only, and then trim the result. Just in case...

Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/business/WorkflowBusiness.java
```diff
@@ -113,6 +115,31 @@ public WorkflowBusiness() {
         }
     }
 
+    private String selectEngineEndpoint(String applicationClass) throws BusinessException {
+        long min = 100000;
+        long tmp;
```
glatard: Variable could be declared at line 126. Could be renamed to, e.g. "runningWorkflows".

Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/business/WorkflowBusiness.java
```diff
@@ -113,6 +115,31 @@ public WorkflowBusiness() {
         }
     }
 
+    private String selectEngineEndpoint(String applicationClass) throws BusinessException {
+        long min = 100000;
```
glatard: Hum...

Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/business/WorkflowBusiness.java
```diff
@@ -113,6 +115,31 @@ public WorkflowBusiness() {
...
+            List<Engine> availableEngines = ApplicationDAOFactory.getDAOFactory().getEngineDAO().getByClass(applicationClass);
+            for (Engine engine : availableEngines) {                
+                tmp = workflowDAO.getNumberOfRunningPerEngine(engine.getEndpoint());
+                if (tmp < min) {
+                    min = tmp;
+                    engineBean = engine;
+                }
+            }
+            endpoint = engineBean == null || engineBean.getEndpoint().isEmpty()
+                    ? Server.getInstance().getMoteurServer()
```
glatard: Is it a fallback? If yes, then we may want to rename getMoteurServer (and all the related variables, including in config file) to FallbackMoteurServer.

Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/client/view/system/classes/EditClassLayout.java
camarasu: OK

Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/business/WorkflowBusiness.java
camarasu: ok

Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/business/WorkflowBusiness.java
camarasu: I'm not sure if we are talking about the same think, but no, for me this is not a fallback.

Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/business/WorkflowBusiness.java
```diff
@@ -113,6 +115,31 @@ public WorkflowBusiness() {
         }
     }
 
+    private String selectEngineEndpoint(String applicationClass) throws BusinessException {
```
glatard: A problem I see with this selection method is that a failing engine may become a black hole receiving all the workflows and crashing them as soon as they are launched. The number of running workflows on such an engine would always be lower than on the other ones. For this reason, random or round-robin selection seems wiser. We should also think of a way to enable/disable engines in the interface (manually), and to automatically disable engines when they obviously fail, e.g. connection fails (workflow cannot be even submitted, we don't get an id back). Exponential backoff seems to be a good solution to enable/disable engines automatically. That could be done by adding a few parameters to the engines table (number of failures since last success, timestamp of last success, timestamp of last attempt).

Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/business/WorkflowBusiness.java
```diff
@@ -195,13 +222,14 @@ public synchronized String launch(User user, List<String> groups,
...
                     + FilenameUtils.getName(version.getLfn()));
 
-            WorkflowExecutionBusiness executionBusiness = new WorkflowExecutionBusiness(applicationClass);
+            String endpoint = selectEngineEndpoint(applicationClass);
+            WorkflowExecutionBusiness executionBusiness = new WorkflowExecutionBusiness(endpoint);
             Workflow workflow = executionBusiness.launch(applicationName,
                     applicationVersion, applicationClass, user, simulationName,
                     workflowPath, parameters);
             if(workflow == null)
                 throw new BusinessException("Workflow is null");
```
glatard: This is where we could increment an error count for the engine.

Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/business/WorkflowExecutionBusiness.java
```diff
@@ -54,11 +56,17 @@
...
-    public WorkflowExecutionBusiness(String applicationClass) throws BusinessException {
+    public WorkflowExecutionBusiness(String engineEndpoint) throws BusinessException {
 
-        engine = WorkflowEngineInstantiator.create(applicationClass);
-    }
+        //HACK for testing while still having simulations launched with VIP 1.16.1; to be removed before getting in production or replaced with a proper constant
+        if(engineEndpoint == null){
+            logger.info("WorkflowExecutionBusiness, endpoint is null, setting it to http://data-manager.grid.creatis.insa-lyon.fr/cgi-bin/m2Server-gasw3.1/moteur_server");
+            engineEndpoint="http://data-manager.grid.creatis.insa-lyon.fr/cgi-bin/m2Server-gasw3.1/moteur_server";
+        }
```
glatard: Isn't it the fallback that was formerly gotten from Server.getInstance().getMoteurServer()? I don't think you need to do this: let's stick to the multiple engine code here.

Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/business/WorkflowExecutionBusiness.java
```diff
@@ -83,7 +91,8 @@ public Workflow launch(String applicationName, String applicationVersion,
 
             return new Workflow(workflowID, user.getFullName(), 
                     engine.getMode().equalsIgnoreCase("pool") ? WorkflowStatus.Queued : WorkflowStatus.Running,
-                    new Date(), null, simulationName, applicationName, applicationVersion, applicationClass);
+                    new Date(), null, simulationName, applicationName, applicationVersion, applicationClass, 
+                    engine.getMode().equalsIgnoreCase("pool") ? "ShiwaPool" : ((WebServiceEngine)engine).getAddressWS());
```
glatard: Feel free to remove any dependence to the ShiwaPool if it carries too much untested extra code.

Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/business/WorkflowBusiness.java
camarasu: Yep, you're right

Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/business/WorkflowExecutionBusiness.java
camarasu: Oh, that's the fallback you were talking about. 
In this case it is, but we probably don't need it. I used it here only for testing purposes (simulations run with the production portal have no engine)

Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/business/WorkflowBusiness.java
camarasu: I'll get back on this in the following days.

Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/dao/mysql/ClassData.java
```diff
@@ -65,26 +65,8 @@ public ClassData() throws DAOException {
...
-        } catch (SQLException ex) {
-            if (ex.getMessage().contains("Duplicate entry")) {
-                throw new DAOException("A class named \"" + appClass.getName() + "\" already exists.");
-            } else {
-                logger.error(ex);
-                throw new DAOException(ex);
-            }
-        }
+        addToClass(appClass.getName(), appClass.getEngines(), "engine");
+        addToClass(appClass.getName(), appClass.getGroups(), "group");
```
glatard: It doesn't see to be necessary to have this done in a method, in particular  given the switch-case statement in the method. I'd just to the DB queries here, as done for the other methods.

Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/dao/mysql/ClassData.java
```diff
@@ -94,24 +76,13 @@ public void add(AppClass appClass) throws DAOException {
      */
     @Override
     public void update(AppClass appClass) throws DAOException {
+        
+        removeFromClass(appClass.getName(), "engine");
+        addToClass(appClass.getName(), appClass.getEngines(), "engine");
```
glatard: Why not an update query?

Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/dao/mysql/ClassData.java
glatard: If this is the only place where these queries are used, I'd avoid using a separate method (in particular in view of the switch-case statement in it).

Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/business/WorkflowBusiness.java
glatard: So, what is returned by Server.getInstance().getMoteurServer()? When is it supposed to be used?

Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/business/WorkflowExecutionBusiness.java
glatard: OK but this code will never be used during testing phase, and it won't be used in production either. I'm wondering if we should have a fallback for these cases where no engine is declared, or if we just want to make sure that this situation doesn't happen (we could raise a proper error message when it does). Thinking about it, I'm not very keen about a fallback.

Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/dao/mysql/ClassData.java
camarasu: I do not think this could be done with an update since we can have a different number of lines (engines per class) at the beginning and at the end of the operation (e.g. go from 1 to 2 or from 2 to 1 engines per class). 
Do you agree ?

glatard: Then, update means either adding engines or removing engines. Why don't we have an add and a remove method that can deal with a list of engines? It sounds like an overkill to remove all the engines and add the new ones.

glatard: OK, let's leave it like that (number of engines is small anyway).

Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/client/view/system/classes/EditClassLayout.java
camarasu: Actually, since split() returns a String[], we can't simply apply a trim() to it. 
We could remove empty spaces like this : split("\\s*,\\s*"). 
Would that be fine with you ?

glatard: Not sure... what happens if there is not space character before or after the comma? Does it still match? I would split by ",", and then trim the result, e.g. as follows:
```Java
Arrays.asList(engines.split(",")).stream().map(String::trim)
```

(Stream was introduced in Java 8)

camarasu: Wow, thanks for the suggestion, I wasn't aware of these new functionalities.:-) And it actually goes a bit further, since the Stream<String> needs to be converted to a String [] : 
this.enginesPickList.setValues(Arrays.asList(engines.split(",")).stream().map(String::trim).toArray(elem -> new String[elem]));

Just a note on my previous message, which was "misinterpreted" by the browser. The argument between " " was actually \\s*,\\s* , so it should have worked fine with and without empty spaces. 

Your suggestion is mush nicer though, so I'll be happy to implement it.:-)

glatard: Well if your suggestion splits and trims at the same time then it's better than mine...

Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/business/WorkflowExecutionBusiness.java
camarasu: It is used by checkRunningSimulations and current simulations (those launched with the current production portal) have no engine attribute (needed in WorkflowExecutionBusiness(simulation.getEngine())).
That's why we need a fallback for testing purposes. The fallback should be replaced with an exception before going into production.

glatard: OK

camarasu:
Pull request updated

glatard:
Nice! Your branch is now in conflict due to the update in version numbers that was hotfixed to branch develop. You should merge develop in your own fork and fix the conflict (using your version numbers).

camarasu:
Conflict solved

PR #31, opened 2016-04-26
Account creation can now be disabled for certain countries

glatard:
@camarasu, please review asap so that we can deploy this hotfix. Note that adding "?w=1" to the URL will show the diff without all the whitespaces that were modified when I re-indented the source code.


camarasu:
Excellent, thank you for this hotfix !

PR #45, opened 2016-06-07
Allow cross-origin requests

glatard:
This is to allow JavaScript pages hosted elsewhere than on VIP to make queries to the API and display information. This is requested by FLI-IAM to use the VIP API in their portal. Ideally, this should be enabled only for the API and not for its authentication method. Currently, it is allowed for any page.


camarasu:
OK

PR #48, opened 2016-06-21
Redmine feature 2838 : a flag type input is managed as a checkbox following by the flag value (command-line-flag in json descriptor)

pascalgirard:
The user can choose to add or not the flag value in the command line

### DIFF
--- a/Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/client/bean/Source.java
+++ b/Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/client/bean/Source.java
@@ -39,7 +39,7 @@ import com.google.gwt.user.client.rpc.IsSerializable;
  */
 public class Source implements IsSerializable {
 
-    private String name, type, userLevel, description, defaultValue,vipTypeRestriction, prettyName ;
+    private String name, type, userLevel, description, defaultValue, vipTypeRestriction, prettyName ;
 
     private boolean optional;
 
--- a/Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/client/bean/Source.java
+++ b/Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/client/bean/Source.java
@@ -39,25 +39,28 @@ import com.google.gwt.user.client.rpc.IsSerializable;
  */
 public class Source implements IsSerializable {
 
-    private String name, type, userLevel, description, defaultValue;
+    private String name, type, userLevel, description, defaultValue,vipTypeRestriction, prettyName ;
+
     private boolean optional;
 
     public Source() {
     }
 
-    public Source(String name, String type, String userLevel, String description, String optional, String defaultValue) {
+    public Source(String name, String type, String userLevel, String description, String optional, String defaultValue, String vipTypeRestriction, String prettyName) {
         this.name = name;
         this.type = type;
         this.userLevel = userLevel;
         this.description = description;
         this.optional = Boolean.parseBoolean(optional);
         this.defaultValue = defaultValue;
+        this.vipTypeRestriction = vipTypeRestriction;
+        this.prettyName = prettyName;
     }
 
     public Source(String name, String type, String userLevel, String description) {
-        this(name, type, userLevel, description, "false","");
+        this(name, type, userLevel, description, "false","","","");
     }
-
+    
     public String getDescription() {
         return description;
     }
@@ -86,6 +89,20 @@ public class Source implements IsSerializable {
         return type;
     }
     
+    public String getVipTypeRestriction() {
+        return vipTypeRestriction;
+    }
+
+    public void setVipTypeRestriction(String vipTypeRestriction) {
+        this.vipTypeRestriction = vipTypeRestriction;
+    }
     
+    public String getPrettyName() {
+        return prettyName;
+    }
+
+    public void setPrettyName(String prettyName) {
+        this.prettyName = prettyName;
+    }
     
 }
--- a/Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/client/view/launch/InputFlagLayout.java
+++ b/Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/client/view/launch/InputFlagLayout.java
@@ -62,8 +62,8 @@ public class InputFlagLayout extends AbstractSourceLayout{
     private CheckboxItem cbFlagInputItem ;
             
     public InputFlagLayout(String name, String comment, boolean optional, String defaultValue, String vipTypeRestriction, String prettyName) {
-        super(name, comment,optional);
-        cbFlagInputItem = new CheckboxItem(prettyName);
+        super(prettyName, comment,optional);
+        cbFlagInputItem = new CheckboxItem(defaultValue);
         cbFlagInputItem.setValue(false);
         cbFlagInputItem.setValueField(defaultValue);
         cbFlagInputItem.addChangeHandler(new ChangeHandler() {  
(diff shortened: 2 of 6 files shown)


Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/client/view/launch/InputFlagLayout.java
```diff
@@ -0,0 +1,101 @@
...
+import fr.insalyon.creatis.vip.application.client.view.common.AbstractSourceLayout;
+import fr.insalyon.creatis.vip.core.client.view.util.FieldUtil;
+import com.smartgwt.client.widgets.form.fields.events.ChangeEvent;  
+import com.smartgwt.client.widgets.form.fields.events.ChangeHandler; 
+
+/**
+ *
+ * @author pgirard
+ * 
+ * Class for display an flag type input as a checkbox following by the flag value (command-line-flag in json descriptor).
```
camarasu: Class for displaying a flag type input as a checkbox followed by ...

Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/client/view/launch/InputFlagLayout.java
```diff
@@ -0,0 +1,101 @@
...
+ * Class for display an flag type input as a checkbox following by the flag value (command-line-flag in json descriptor). 
+ * With this checkbox the user can add or not the flag value on the command line.
+ */
+public class InputFlagLayout extends AbstractSourceLayout{
+    
+    private CheckboxItem cbFlagInputItem ;
+            
+    public InputFlagLayout(String name, String comment, boolean optional, String defaultValue, String vipTypeRestriction, String prettyName) {
+        super(name, comment,optional);
+        cbFlagInputItem = new CheckboxItem(prettyName);
```
camarasu: I think it would be more useful to have the flag value instead of its prettyName

Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/client/bean/Source.java
```diff
@@ -39,25 +39,28 @@
  */
 public class Source implements IsSerializable {
 
-    private String name, type, userLevel, description, defaultValue;
+    private String name, type, userLevel, description, defaultValue,vipTypeRestriction, prettyName ;
```
glatard: space missing before "vipTypeRestriction".

Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/client/view/launch/InputFlagLayout.java
```diff
@@ -0,0 +1,101 @@
...
+        cbFlagInputItem.setValue(false);
```
glatard: Shouldn't the value depend on defaultValue?

Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/client/view/launch/LaunchTab.java
```diff
@@ -113,7 +112,7 @@ public void onSuccess(Descriptor descriptor) {
                     }
                 }
                 mandatorySources.addAll(optionalSources);
-                
+
```
glatard: Remove whitespaces.

Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/client/view/launch/LaunchTab.java
```diff
@@ -141,7 +147,7 @@ public void onSuccess(Descriptor descriptor) {
                 modal.hide();
 
                 configureInputsLayout(true);
-
+
```
glatard: Remove whitespaces.

Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/business/simulation/parser/GwendiaParser.java
```diff
@@ -56,7 +56,7 @@ public GwendiaParser() {
     public void startElement(String uri, String localName, String qName, Attributes attributes) throws SAXException {
 
         if (localName.equals("source")) {
-            Source s = new Source(attributes.getValue("name"),attributes.getValue("type"),attributes.getValue("user-level"),attributes.getValue("description"),attributes.getValue("optional"),attributes.getValue("default"));
+            Source s = new Source(attributes.getValue("name"),attributes.getValue("type"),attributes.getValue("user-level"),attributes.getValue("description"),attributes.getValue("optional"),attributes.getValue("default"), attributes.getValue("vip-type-restriction"),attributes.getValue("pretty-name"));
```
glatard: Such long argument lists could be split in multiple lines.

glatard:
Good work! See (minor) comments above. Also, why isn't prettyName used to display the inputs in the launch form?

Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/client/view/launch/InputFlagLayout.java
glatard: The pretty name is meant to be used for *all* inputs, and displayed in the launch form instead of the input id.

camarasu: Indeed, but I think we are not talking about the same piece of text. Here the prettyName concerns the text on the right hand-side of the checkbox. The label above the checkbox should, indeed, display the pretty name as its is meant to be for all inputs.

pascalgirard:
Thank you for your remarks. I hope this commit responds to them. 
During these code correction and commit, I didn't see whitespaces. I hope the commit solve the problem. 
For the moment, I put the pretty-name just for the flag checkbox, but I know that It has to be displayed for all fields in the launch tab. For others inputs it needs to modify the gwendia.vm template again and other files, and test. But I'm wondering if adding a checkbox for optional input is more a priority before my holidays. In any case I will did it in redmine feature 2803. It is as you like.
In the redmine 2803 feature I will work on the display of inputs and output like the screenshot of cbrain Tristan put in slack.

Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/client/view/launch/InputFlagLayout.java
```diff
@@ -62,8 +62,8 @@ Sorina Pop (sorina.pop@creatis.insa-lyon.fr)
     private CheckboxItem cbFlagInputItem ;
             
     public InputFlagLayout(String name, String comment, boolean optional, String defaultValue, String vipTypeRestriction, String prettyName) {
-        super(name, comment,optional);
-        cbFlagInputItem = new CheckboxItem(prettyName);
+        super(prettyName, comment,optional);
```
glatard: I'm not sure if that works. Be careful, when the launch button is hit, the *id* must be passed in the input document, not the pretty name. The pretty name may be used only in the GUI. I'm not sure how AbstractSourceLayout uses the name, make sure it is not used for anything else than the GUI.

pascalgirard:
I think you're right. The launch button launches the application but the execution is killed just after. I'm going to investigate that and find another solution.

glatard:
yes, here is what the workflow execution says: ```23/06/16 16:05> Fatal error: dataset read error: fr.cnrs.i3s.moteur2.exceptions.DataSetException: Source "Flag for test" not found in workflow```

pascalgirard:
Ok I came back to super(name,...), and I'm looking for another solution for display the pretty-name.

pascalgirard:
In fact the input name on the launch tab is constructed in the parent constructor of InputLayout, GateLabSourceLayout and InputFlagLayout classes (in AbstractSourceLayout class).
To avoid modifying the parent constructor signature, so to avoid an impact on InputLayout and GateLabSourceLayout classes, I did a temporaly and bad thing. In AbstractSourceLayout class I overloaded the constructor with the prettyName parameter. This new constructor is currently used by InputFlagLayout class only.
In the redmine feature 2803 I will delete the constructor without the prettyName parameter. I note this in AbstractSourceLayout class. I hope this temporaly and bag solution is acceptable for few days.

glatard:
OK, let's talk about this, I'm a bit confused now.

PR #49, opened 2016-06-24
Launch form gui improvements

pascalgirard:
Redmine feature 2803 : Launch form GUI improvements.
First item of this Redmine feature : checkbox to activate/disable optional inputs.
It works but to to complete this first item, we have to realise redmine feature 2980 we saw today with Sorina.
It is a proposition of display the checkbox. It can be modify.
checkbox


glatard:
Nice one! I think we could still improve a bit the GUI though:
* For flags: the id and name are redundant. It looks confusing. We should keep only the name, not the id. Put the name in bold. 
* For optional inputs: the label "activate this optional field" is not necessary. The checkbox could be put before the input id. When the checkbox is not ticked, the input should be greyed out. Typing a string in the text field (or adding a file) should activate the checkbox.

pascalgirard:
- For flags : you're right and it's done. I put a wrong screenshot in this discussion. Sorry for that. Below, here is the correct screenshot, with the name of the flag and its value "-v".
- For optional inputs : ok for the label. Ok to put the checkbox before the input id. It seems to me that we talked about the pretty-name instead of the id, did we ? For all inputs/output (redmine 2837) ? "Typing a string in the text field (or adding a file) should activate the checkbox." : I will try that.
correct_screenshot_for_flag

glatard:
- put the flag checkbox on the same line than the flag (pretty name).
- remove label "activate this optional field".
- use pretty name everywhere.

glatard:
Great! Please post an updated screenshot whenever it makes sense.

pascalgirard:
I will add comments tomorrow. Please see 2 screenshots
pr_2803_1
pr_2803_2

glatard:
The screenshots look very good. The second one is a workflow that contains comments that are no longer necessary and now confusing (e.g. "put 'no' to ignore this optional input). Make sure that these comments are not added by the Boutiques importer (in gwendia template). If they are, remove them accordingly. Similarly, the blue text in the second screenshot could be omitted (i.e. removed from the gwendia boutiques template).

Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/client/view/common/AbstractSourceLayout.java
```diff
@@ -1,121 +1,156 @@
...
-import com.smartgwt.client.widgets.Label;
-import com.smartgwt.client.widgets.layout.HLayout;
-import com.smartgwt.client.widgets.layout.VLayout;
-import fr.insalyon.creatis.vip.core.client.view.util.WidgetUtil;
-
-/**
- *
- * @author Rafael Ferreira da Silva
- */
-public abstract class AbstractSourceLayout extends VLayout {
```
glatard: Somehow, the whole file looks deleted and re-created in git. Try to avoid that as it masks your edits.

Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/client/view/common/AbstractSourceLayout.java
```diff
@@ -1,121 +1,156 @@
...
+    protected HLayout sourceCommentHLayout;
+    protected HLayout sourceFieldHLayout;
+    protected HLayout fieldHLayout;
+    
+    protected HLayout flagCbHLayout; // Layout in which a checkbox (flag input) can be add. This layout is on the left of the sourceLabelHLayout.
+    protected LayoutSpacer sourceCommentLayoutSpacer; // Empty layout on the left of the sourceCommentHLayout, to horizontaly align it with sourceLabelHLayout and sourceFieldHLayout.
+    protected boolean optional;
+
+    /**
+     * Currently GateLabSourceLayout object use it.
```
glatard: Not really a good comment to describe the method.

pascalgirard:
1) I confirm that the new gwendia template (gwendia.wm) doesn't contain comments you are talking about : the text in blue or in green or sentences like "put 'no' to ignore this optional input".
So the launch form of new applications will be displayed like the screenshot of "optional_input_2803 1" above.
But for applications, already imported, theses textes are still displayed because they are in the "application_name.gwendia" of each application. To avoid this, we will have to modify all "  .gwendia".

2) With Sorina we discussed about displaying of optional inputs. We propose you the points below :
2.1) when the launch form is displayed, just after the user clicks on his application, the checkbox will be unchecked, the optional input will be disabled and the list or range values will be "No_value_provided". The user sees "No_value_provided", but he can't modify it because the input is disabled.
2.2) if the user does not click on the checkbox to activate the input, ""No_value_provided" value is sent to the workflow for list or range fields.
2.3) if the user clicks on the checkbox, the input is activate and the "No_value_provided" string disapear, to avoid for the user to delete it.
2.4) if the user doesn't fill the field and click on "Launch" button, the screen will display the icon for empty fields (please see image below) with a text adapted for optional inputs : something like "This field is required, because you selected it. If you don't need this field, please deselect it."
 
Do you agree with theses propositions ?

optional_input_tool_tip

pascalgirard:
In fact the 2.4) point is in the case of the user clicked on the checkbox before, to activate the input. So the 2.4) point is a sub point of 2.3)

glatard:
Yes to all! In principle there is a slight difference between not activating an optional input and passing an empty string. But since VIP doesn't allow for empty strings as input parameters, this detail is not important.

pascalgirard:
All points of item 2) (comment 2016/12/09) will be implements in the new PR.

PR #50, opened 2016-07-07
Develop appimporter

fredericcervenansky:
wrapper for msseg challenge is working.
More, a correction to allow user's email as input.


glatard:
Wonderful! Minor comment: The user email could be extracted from the JSON descriptor and added to the workflow as a string constant. That would prevent us from copy-pasting it anytime we launch the pipeline in test phase (with risk of error). Do you think you'd have time to do this, or do we live with it for now?

fredericcervenansky:
I could try to implement it but the parsing of descriptors is generic. Do we really need these tags for all applications (not only for the challenge)?

glatard:
No, that would be specific to the challenge. It's in our doc:
`
These following attributes are required.
"vip:miccai-challenger-email"
"vip:miccai-challenge-team-name"
`
The regular template wouldn't use these attributes (they will be null in the bean object).

fredericcervenansky:
Yep but the parsing of json files for these tags will be performed for every template (challenge or not).

glatard:
Exactly, and if we don't find them, we just use a null value. I don't really see any problem with that, do you?

fredericcervenansky:
Add automatic parsing of challenger email and team during workflow deployement

PR #52, opened 2016-07-11
Feature/api test refactoring

axlbonnet:
In order to make the API business classes testable, I made the following changes :
- Api business do not create dependencies on demand with "new". These dependencies are now injected in constructor.
- Common API business is not done in common superclass constructor, but in a separate dependency. A class ApiContext is created to contain the per-request API information (http stuff, user, warnings) and each API method take it in input.

To illustrate the goal of these changes, I created some unit test where dependencies are mocked (with the mockito library) which demonstrate the bug #2983. The failing test is commented to not block the maven build.

### DIFF
--- a/Portal/vip-api/pom.xml
+++ b/Portal/vip-api/pom.xml
@@ -32,7 +32,8 @@ knowledge of the CeCILL-B license and that you accept its terms.
 
 -->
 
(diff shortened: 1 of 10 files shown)


glatard:
This looks excellent! Unit testing is really missing and required. Thanks for this nice contribution.

PR #54, opened 2016-07-13
Bugfix/2983 get execution

axlbonnet:
Correct bug and make unit test OK
The bug concern the permission of non-admin users on execution related SOAP APIs.


glatard:
Nice job @axlbonnet!

PR #59, opened 2016-10-13
add output name and filename in CARMIN execution results links

axlbonnet:
Feature #3022

The download links returned by the CARMIN API contain 2 more parameters :
- outputname : the name of the output parameter
- filename : the filename (not the path)

For instance : `https://vip.creatis.insa-lyon.fr/api/../fr.insalyon.creatis.vip.portal.Main/filedownloadservice?operationid=Download-7367171614511&outputname=final_file&filename=final_file.txt`

Beware that in a SOAP response (XML), the '&' char is encoded which give

```
<returnedValueListStrings>
   <item>https://vip.creatis.insa-lyon.fr/api/../fr.insalyon.creatis.vip.portal.Main/filedownloadservice?operationid=Download-7367171614511&amp;outputname=final_file&amp;filename=final_file.txt</item>
</returnedValueListStrings>
```

### DIFF
--- a/Portal/vip-api/src/main/java/fr/insalyon/creatis/vip/api/business/ExecutionBusiness.java
+++ b/Portal/vip-api/src/main/java/fr/insalyon/creatis/vip/api/business/ExecutionBusiness.java
@@ -54,14 +54,17 @@ import fr.insalyon.creatis.vip.core.client.bean.User;
 import fr.insalyon.creatis.vip.core.server.business.BusinessException;
 import fr.insalyon.creatis.vip.core.server.business.ConfigurationBusiness;
 import fr.insalyon.creatis.vip.datamanager.server.business.TransferPoolBusiness;
-import java.net.MalformedURLException;
-import java.net.URL;
+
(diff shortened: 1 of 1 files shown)


glatard:
@camarasu I leave it up to you to merge this.

axlbonnet:
@glatard 

> When this method is used to download files independently from any execution, the output name couldn't be used because it may not exist

I don't get exactly what you mean. Do you talk about the fact that it is possible to download files from the grid not linked with an execution ? I agree but i'm quite sure the CARMIN method getExecutionResults can only be applied to an execution. The former links (still used in the portal) will still work as the operationid is the only mandatory parameter in the request.
Or do you mean that it is possible to have execution result without output name? In this case that could be a problem.

glatard:
I mean it is possible to download files from the grid not linked with an execution. There is no problem.

axlbonnet:
FLI-IAM is OK with this functionality in the current state.

I've just tested it again and it's OK, there is just the '&' character encoding to be aware of.

PR #60, opened 2016-10-14
remove mozilla persona auth related stuff

axlbonnet:
Mozilla Persona will shutdown on November 30th. So the persona authentication in VIP should be removed (cf #3024).


Portal/vip-core/src/main/java/fr/insalyon/creatis/vip/core/client/CoreModule.java
```diff
@@ -93,7 +93,9 @@ public void load() {
         // Home Tab
         Layout.getInstance().addTab(homeTab);
 
-        // For users with no group (e.g. they just signed in using Mozilla Persona)
+        // open account tab to accept the terms of use if necessary
+        // Also open the tab for users with no group :
+        // It was possible with former Mozilla Persona but keep it just in case
```
glatard: Just FYI, it is still possible in case of SAML authentication.

PR #61, opened 2016-11-07
Feature/rest api

axlbonnet:
REST API conform to CARMIN 0.2 (and its swagger specification).

Done with Spring MVC (and spring security). Based on the existing SOAP implementation of CARMIN 0.2.

Secured via an api key (in an header) that can be generated and accessed in the account page in the VIP portal.


Portal/vip-api/src/main/java/fr/insalyon/creatis/vip/api/VipConfigurer.java
```diff
@@ -0,0 +1,98 @@
...
+/**
+ * Used to configure vip on startup and daily
+ *
+ * On startup, test database
+ * Daily, renew the grida proxy
+ *
+ * Created by abonnet on 7/26/16.
+ */
+@Component
+public class VipConfigurer implements ApplicationListener<ContextRefreshedEvent> {
```
glatard: Awesome! In the future this part could go in a separate PR though.

PR #62, opened 2016-11-08
reset failed auth count after an auth success

axlbonnet:
On SAML authentication success, reset the number of failed authentications to avoid account locks.


camarasu:
Looks good to me. 
@glatard: You may want to check it and merge if it's fine with you, too.

PR #63, opened 2016-11-08
remove useless gwt maven builds

axlbonnet:
Make the gwt compilation only once, when the vip-portal war is created.

It gains a lot of time (1 min 50 vs 6 min 20 for a complete build on ma machine).


camarasu:
That's really nice !
OK with me to merge the PR.

PR #64, opened 2016-11-10
dont send group mail to locked users

axlbonnet:
Ticket 2972.

Do not include locked users in the following cases :
- mail send to all users
- mail send to a specific group in VIP portal

In particular, it doesn't (yet) concern mails sent to the vip-support group from :
- the portal to report an issue
- VIP itself in case of problem

I didn't thought it necessary because the group vip-support is well managed and, for the second case, it concerns several places in the code base. But I can work on that if necessary.


camarasu:
I agree. 
OK with me to merge the PR.

PR #65, opened 2016-11-18
send 'kill with reason' messages to vip-support in copy

axlbonnet:
cf redmine #3038


camarasu:
Seems good to me, thank you.

PR #68, opened 2016-12-14
Launch form ranges

pascalgirard:
2 comments :
1) validate() method from LaunchFormLayout class :
Removal of code below 

if (source.isOptional() && (source.getValue() == null || source.getValue().equals("") || source.getValue().equals("null"))) {
	source.setValue("No_value_provided");
} else

because optional input (=source here) are managed in InputLayout.java (constructor and in configureOptionalInputCheckbox() method).

2) I added 2 new constants in ApplicationConstants.java
public static final String INPUT_WITHOUT_VALUE = "No_value_provided";
public static final String INPUT_WITHOUT_VALUE_REQUIRED_MESSAGE = "Either you deselect the field or you fill it";

I know that it is probably not the right place.

From my experience, all web applications I saw, had a file to gather and centralize all GUI textes.
In this file, each text was assigned to a variable, and it was this variable that was used (possibly several times) in GUI code.
This is the topic of redmine feature 2977 (GWT provides a API for that).
In the meantime, to gather some new texts before implementation of 2977, I propose to 
- either create a java class (like ApplicationConstants.java) available for all VIP modules,
- or put texts in a separate section in the ApplicationConstants.java file.
What do you think ?


pascalgirard:
Please click on http://vip.creatis.insa-lyon.fr:7070/vip-portal-1.19_1/ to test this PR.

pascalgirard:
A little screenshot
pr_ranges

pascalgirard:
When the user launches his application, optional field are disabled
form_launch_display_by_default

After selecting the optional inputs

form_launch_selection

I tried to launch the application without values. The required message is adapted to optional inputs

form_launch_without_values

I filled in some fields, the required message disappeared for fields ok.

form_launch_some_values_filled_in

If I deselect fields all values are deleted and the value "No_provided_value" is passed. Please see the first screenshot.

PR #69, opened 2016-12-16
Bugfix/#3026 session cookie

axlbonnet:
do not change session ID at each request to avoid conflict and authentication errors


Portal/vip-api/src/main/java/fr/insalyon/creatis/vip/api/business/AuthenticationBusiness.java
```diff
@@ -62,7 +62,6 @@ public void authenticateSession(String userName, String password) throws ApiExce
              User user = configurationBusiness.signin(userName, password);
              
              AbstractAuthenticationService.setVIPSession(apiContext.getRequest(), apiContext.getResponse(), user);
-             configurationBusiness.updateUserLastLogin(userName);
```
glatard: Well this is quite useful to know when people logged in for the last time.

Portal/vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/business/ConfigurationBusiness.java
```diff
@@ -106,8 +106,6 @@ public boolean validateSession(String email, String session) throws BusinessExce
             if (email != null && session != null) {
                 UserDAO userDAO = CoreDAOFactory.getDAOFactory().getUserDAO();
                 if (userDAO.verifySession(email, session) && !userDAO.isLocked(email)) {
-                    String newSession = UUID.randomUUID().toString();
-                    userDAO.updateSession(email, newSession);
```
glatard: The reason for this is that session ids can only be used once, therefore if someone intercepts the session id he/she couldn't use it to log in again. What is the motivation for removing it?

Portal/vip-api/src/main/java/fr/insalyon/creatis/vip/api/business/AuthenticationBusiness.java
axlbonnet: This is just a duplicate removal, the same method is called in setVIPSession (the line before).

Portal/vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/business/ConfigurationBusiness.java
axlbonnet: In my knowledge session ids can be used than once. The standard use is that the session id is unique and doesn't change for a single session (from login to logout).
In fact it is not done at each request in VIP, but only at each load (in each F5 in not in each AJAX request).

The best solutions to avoid session hijacking is to use SSL/TLS (done), and to regenerate the session id after the login (done in VIP).

To change the session id with each request is an other way to reinforce the security but is known to have side effects because concurrent calls can lead to session invalidation. This is exactly what happened in VIP in SOAP requests. Closely timed requests cause session id jam and authentication errors. Then after 5 requests the account is locked which is a major problem. This happened in selenium automatic tests.

Session id regeneration should be used along with a mechanism to avoid this concurrency problem. For instance remembering for a short period the old cookies but that implies some work.
I think it does not bring much value in VIP but has a major side-effect and it is best to remove it.

Some resources : 
- https://en.wikipedia.org/wiki/Session_hijacking#Prevention
- http://php.net/session_regenerate_id (PHP implementation, see warning)

glatard: Thanks for the note, I agree!

PR #71, opened 2017-01-06
Execution name star

pascalgirard:
Redmine 2932 : Application launch form : red star on "Execution name" because this field is mandatory.


pascalgirard:
This PR will be reopened from branch develop and not from branch launch_form_ranges.

PR #73, opened 2017-01-10
New VIP version : tag 1.20

pascalgirard:


camarasu:
It seems to me there are three pom files missing : warehouse, query and cardiac. Could you please update the version numbers for them also ?

camarasu:
OK, true, those modules are no longer part of the VIP module list.

PR #79, opened 2017-02-23
Fix missing dependency definitions

vincent-legoll:
This PR adds a bunch of missing dependency definitions in pom.xml files

They were required to "mvn compile" in Portal subfolder

### DIFF
--- a/Portal/vip-application-importer/pom.xml
+++ b/Portal/vip-application-importer/pom.xml
@@ -95,6 +95,12 @@ knowledge of the CeCILL-B license and that you accept its terms.
         </dependency>
      
         <dependency>
+            <groupId>org.apache</groupId>
+            <artifactId>velocity-tools</artifactId>
+            <version>2.0</version>
+        </dependency>
+       
+        <dependency>
             <groupId>org.codehaus.plexus</groupId>
             <artifactId>plexus-archiver</artifactId>
             <version>2.4.4</version>
--- a/Portal/vip-application/pom.xml
+++ b/Portal/vip-application/pom.xml
@@ -131,6 +131,30 @@ knowledge of the CeCILL-B license and that you accept its terms.
         </dependency>
 
         <dependency>
+            <groupId>org.jivesoftware</groupId>
+            <artifactId>smack</artifactId>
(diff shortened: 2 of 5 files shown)


glatard:
All looks good to me, thanks!

camarasu:
Vincent, could you please tell us more about why you needed to add these dependencies?
As far as I can see, they are included in a transitive manner due to other dependencies that depend on them. Here is an example: grida-common is a dependency of grida-client, which is a dependency of vip-core, which is a dependency of most VIP sub-projects.
In my opinion, adding them a second time (and with possibly different versions) is not optimal.

vincent-legoll:
I'm not sure.

I tried mvn compile and it failed with msgs like this:

[ERROR] Failed to execute goal on project vip-core: Could not resolve dependencies for project fr.insalyon.creatis:vip-core:jar:1.20: The following artifacts could not be resolved: fr.insalyon.creatis:grida-client:jar:1.4, fr.insalyon.creatis:devtools:jar:0.1, com.smartgwt:smartgwt:jar:3.1: Could not find artifact fr.insalyon.creatis:grida-client:jar:1.4 in JBoss (https://repository.jboss.org/nexus/content/groups/public) -> [Help 1]

To fix that, I did : 

added a repo to VIP-portal/pom.xml:
        <repository>
          <id>dependencies</id>
          <url>file://${basedir}/../dependencies</url>
        </repository>

then 

mvn install:install-file -Dpackaging=jar -DlocalRepositoryPath=dependencies -DgroupId=com.smartgwt -DartifactId=smartgwt -Dversion=3.1 -Dfile=/home/vip/apache-tomcat-8.5.11/webapps/vip-portal-1.20/WEB-INF/lib/smartgwt-3.1.jar

And repeated until everything compiled...

This is maybe the wrong way to fix it...

camarasu:
OK, as I understand it, the problem was that the dependencies (that were correctly declared in the pom.xml), were not found by Maven. This is "normal" for some of them, since our Nexus is not up to date. In this case, you need to install them locally, but you shouldn't need to add them in the pom.

PR #80, opened 2017-02-23
Add new CoreConstants for apache.ssl.port & apache.host vip.conf entries

vincent-legoll:
Those 2 were alone & harcoded, move them alongside the other ones


camarasu:
Looks good to me, thank you !

PR #81, opened 2017-03-09
Move away from SSLv3 to TLS

vincent-legoll:
- Following SSL V3.0 "Poodle" Vulnerability - CVE-2014-3566
  SSLv3 has been disabled by default in JDKs (8u31, 7u75, 6u91)
  See: https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2014-3566
- Change MyProxy SSL socket settings to use TLS instead

Discovered as the following error message was logged:
javax.net.ssl.SSLHandshakeException: No appropriate protocol
(protocol is disabled or cipher suites are inappropriate)

### DIFF
--- a/Portal/vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/business/proxy/ProxyClient.java
+++ b/Portal/vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/business/proxy/ProxyClient.java
@@ -188,7 +188,7 @@ public class ProxyClient {
 
     private void connect() throws Exception {
 
-        SSLContext sc = SSLContext.getInstance("SSL");
+        SSLContext sc = SSLContext.getInstance("TLS");
         sc.init(new KeyManager[]{}, new TrustManager[]{new ProxyClient.MyTrustManager()}, new SecureRandom());
         SSLSocketFactory sf = sc.getSocketFactory();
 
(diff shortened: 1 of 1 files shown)


vincent-legoll:
This change allowed us to get past the SSLHandshakeException.

Please review from a security point of view.

Do we want to allow  all those TLS versions ?

glatard:
Looks good to me!

Portal/vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/business/proxy/ProxyClient.java
```diff
@@ -188,15 +188,15 @@ private void addVomsExtension(String vo) throws Exception {
...
+        SSLContext sc = SSLContext.getInstance("TLS");
         sc.init(new KeyManager[]{}, new TrustManager[]{new ProxyClient.MyTrustManager()}, new SecureRandom());
         SSLSocketFactory sf = sc.getSocketFactory();
 
         this.socket = (SSLSocket) sf.createSocket(
                 Server.getInstance().getMyProxyHost(),
                 Server.getInstance().getMyProxyPort());
 
-        this.socket.setEnabledProtocols(new String[]{"SSLv3"});
+        this.socket.setEnabledProtocols(new String[]{"TLSv1", "TLSv1.1", "TLSv1.2"});
```
camarasu: We could only keep v1.2

camarasu:
Looks good to me too, but I suggest we only keep TLS v1.2. 
I tested it and works fine.

vincent-legoll:
Updated to only use TLS v1.2

PR #93, opened 2017-03-21
Fix undesired email domain handling

vincent-legoll:
An undesired domain of, for example, "t.com" was matching
against "user@hit.com", which is counter-intuitive.


glatard:
Looks good to me!

PR #120, opened 2017-04-11
3023 bugfix get execution status

axlbonnet:
When getting a single execution in the Carmin API, it is necessary to update this execution's status if it's running to be informed of its completion.


camarasu:
Looks good to me, thank you for the fix!

glatard:
Looks good!

PR #126, opened 2017-05-05
Fix bug when email handling lists are empty

vincent-legoll:
- Undesired country & domain lists in config file can be empty. But the
  org.apache.commons.configuration.PropertiesConfiguration.getList() API
  returns a List<String> containing a single empty string in that case.
  So skip those "almost-empty" items.

- Fix some whitespace in the vicinity


camarasu:
Looks good to me.

PR #130, opened 2017-05-29
Gitattribute file : wrapper.vm eol=lf

pascalgirard:
The .gitattributes file to give attributes to pathnames. For the time being, it concerns "wrapper.wm" file. It ensures that "wrapper.wm" file line endings are converted to LF in the repository.


camarasu:
Perfect, thank you. 
Maybe this could be generalized to all *.vm files?

camarasu:
Great, thank you!

PR #133, opened 2017-07-24
chore: fix pom parent version number to 1.21 for 5 modules.

wassong-iphc:


camarasu:
Hi Pascal, thank you for the fix.

PR #138, opened 2017-09-26
Feature/3124 add local or grid option

wassong-iphc:


camarasu:
That's great, thank you for implementing this feature. Looks good to me. 
There are a few lines that must have been automatically re-formated and appear as modified because the extra space at the end of the line was deleted. I think it would be better to keep them in their initial format.

wassong-iphc:
Yes, my editor automatically removed trailing whitespaces.  It is possible to ignore whitespaces in diffs, so that they are more clean and easier to read : add the option "w=1" to the url when viewing the diff.  For more details on this github issue, see https://github.com/camsong/chrome-github-mate/issues/5

camarasu:
OK, thank you!

PR #141, opened 2017-11-10
Bosh integration : WARNING : please read PR comment

pascalgirard:
- This PR allows to create a "json" folder in application file folder of VIP File Transfer. So, an application import creates 4 folders : bin, gasw, json and workflow.
- This PR allows to copy the application json descriptor in the "json" folder.
- The copied json descriptor is the json descriptor used to import the application.
- The a ".tar.gz" file is created in the "json" folder with the copied json descriptor.
- The copied json is available for moteur.
- Moteur can use this file with bosh on newk server. Please see below for an exemple.
- This PR does not allow to produce the json file containing the parameters and their value to use bosh like this : "bosh -i input_param_file.json -e ./tool.json". 
-- In that case bosh is able to launch the application (docker image and the tool inside) => no need of the application wrapper bin.
-- for example : for one execution of Anima_pipeline_test_bosh application the input_param_file.json will be :
{
    "image_to_correct" : "./01005_T1.nii.gz",
    "B_flag_value" : "0.3",
    "image_corrected_name" : "./01005_T1_result.nii.gz"
}

- Python 2.7 (or other) is needed to run bosh. On moteur1 and moteur2 the python version is 2.6. That's why moteur1 needs currently to use bosh on newk server (python 2.7)


pascalgirard:
Be carefull
- generated input_json_parameter.json is removed at the end of the job.
- this code was not tested with several jobs
- this is not executable in moteur1 and 2 because of python version. It works on newk.
- after import, it is necessary to modifie application json descriptor on the grid to follow boutiques schema V5
- application json descriptor does not have indentation and tabulations

PR #144, opened 2017-12-01
Parametrable vo

wassong-iphc:
The VO to use is defined in the configuration file vip.conf.
2 variables must be set, vo.name and vo.root. For example:
vo.name=biomed
vo.root=/grid/biomed


camarasu:
Looks great to me, thank you !

PR #145, opened 2017-12-14
improve groups path parsing

axlbonnet:
When a group name starts with the name of an other group, there is a bug in the parsing.


camarasu:
Looks perfect to me. Thanks for solving this issue!

PR #148, opened 2017-12-22
Replaced DM and GateLab Applets with a solution based on HTML5 + JS 

camarasu:
Note: webkitdirectory works with Chrome, Firefox and Edge, but is not supported in Internet Explorer


axlbonnet:
Seems fine to me !

PR #149, opened 2018-01-12
fix: close proxy file after reading it.

wassong-iphc:
When reading the proxy file in method ProxyClient.getProxy(), close the FileInputStream after reading the content.  This is done using a try with resources.
Also made some minor refactoring in the same method.


camarasu:
Looks great to me, thank you for fixing this!

PR #152, opened 2018-03-09
Improve logs

axlbonnet:
Feature #3179

Change log4j initialization to log everything from the server starttup.

**Do not forget to add "-Dlog4j.configuration=log4j.properties” in tomcat initialization**


camarasu:
That's great, thank you!

PR #163, opened 2018-10-05
Feature/authorize colon character as input

axlbonnet:
Only need to allow the colon character in 2 regex.


camarasu:
Great, thanks !

PR #168, opened 2019-03-01
Feature/openaire

axlbonnet:
Add a doi and a json lfn to all applications
Applications can be published using bosh to get a doi.


camarasu:
That's great, thank you!

PR #169, opened 2019-03-01
Handle array inputs in the API

axlbonnet:
Improve the carmin API :
- accept array as inputs to parallelize jobs
- correct an output bug in getExecutionResults


camarasu:
Great, thanks  !

Issue #175, opened 2019-03-28
Be able to change its email address in the account page


axlbonnet:
Give the possibility to a user to change its own email.

axlbonnet:
Inlcuded in release 1.25

PR #177, opened 2019-03-28
Feature/update email

axlbonnet:
Treats #175 

This consists in :
- update the db schema to cascade the email change from the VIPUsers table
- add the server-side code to allow to change email
- add an user interface in the account page to let an user update its email
- be careful about the web session and cookies

Note :
- I did not add a user interface in the admin page as I didn't find it very useful
- The email address's of the publication owner must me handled separately as the VIP table use MyIsam and does not support foreign keys


Portal/vip-core/src/main/java/fr/insalyon/creatis/vip/core/client/view/user/account/RequestNewEmailLayout.java
```diff
@@ -0,0 +1,109 @@
...
+
+                                @Override
+                                public void onSuccess(User result) {
+                                    Modules.getInstance().userUpdated(CoreModule.user, result);
+                                    CoreModule.user = result;
+
+                                    WidgetUtil.resetIButton(saveButton, "Change email", CoreConstants.ICON_SAVED);
+                                    Layout.getInstance().setNoticeMessage(
+                                            "Your email change has been taken into account. "
+                                                    +" You will receive a mail with a code to validate it.<br />"
```
camarasu: Here is a reformulation suggestion: "Please use the validation code sent to your new email address in order to validate it. Note that your login email will not change until you validate the new address."
I think the refresh info should go into the email, with the activation code.

Portal/vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/business/ConfigurationBusiness.java
```diff
@@ -930,6 +953,26 @@ public void updateUserPassword(String email, String currentPassword,
         }
     }
 
+    public void updateUserEmail(String oldEmail, String newEmail) throws BusinessException {
+        verifyEmail(newEmail);
+        try {
+            CoreDAOFactory.getDAOFactory().getUserDAO().updateEmail(oldEmail, newEmail);
+            // need to update publication separately as the table does not
+            // support foreign keys
+            updatePublicationOwner(oldEmail, newEmail);
```
camarasu: What happens if the email is updated but the updatePublicationOwner fails? Admins should probably have a notification on it.

camarasu:
Great, thank you!

PR #186, opened 2019-04-25
Migrate gwt

wassong-iphc:
- Upgrade to latest version of GWT : 2.8.2
- Upgrade to latest version of smartgwt : 12.0p
- Upgrade to latest version of maven gwt plugin, replacing the legacy plugin by the new recommended one.
- Correction of the problems that appeared due to the upgrades.
- Remove the code of the applets.  They are unused.
- Remove hack to make the scrollbars work for older version of smartgwt of firefox.
- Clean pom.xml files : remove scm/svn info, release plugin, antrun plugin, compilation configuration only in  arent pom, set utf8 encoding on every module.


camarasu:
Thank you, Pascal! Looks great to me.

axlbonnet:
Great work indeed !

Issue #189, opened 2019-05-13
Account update causes problematic DFC folder change


axlbonnet:
First issue : it does not sanitize first name and surname, accents and special characters cause bugs.

2nd issue : With DFC, it is not possible to rename a folder, so it is not possible to change an user's folder

axlbonnet:
Included in release 1.25

PR #190, opened 2019-05-13
Do not update LFC folder on user update

axlbonnet:
Correct #189 

When a user update personal information (organization, phone number), its folder is not update anymore.


camarasu:
Great, thanks for the fix!

PR #193, opened 2019-06-11
change font size for application names and improve their format

axlbonnet:
With the new gwt version, style has changed for text in a tile grid component, and they are too small.
Change the their style for the "normal" style that seems to be the defaut one outside the tile grid.

Also correct the application formatting for long application names.


axlbonnet:
Also correct a bug : the application from a class name with a space could not be select

PR #197, opened 2019-06-27
Visualisation image

wassong-iphc:
Add visualization of images in format : dicom, nifti, mhd/raw and more.


Portal/vip-visualization/src/main/java/fr/insalyon/creatis/vip/visualization/client/view/AbstractViewTab.java
```diff
@@ -62,20 +59,22 @@ public void load() {
 
     private void loadLFN(String lfn) {
         VisualizationServiceAsync vs = VisualizationService.Util.getInstance();
-        modal.show("Loading data file...", true);
+        Layout.getInstance().setNoticeMessage(
```
camarasu: Shouldn't this message disappear (be hidden at some point) ?

Portal/vip-portal/src/main/webapp/js/ami-viewer.js
```diff
@@ -2,23 +2,28 @@
 // https://codesandbox.io/s/github/FNNDSC/ami/tree/master/lessons/03
 
 function amiViewer(url, rawExtension, divId) {
+  window.setNoticeMessage(
+    'Loading and parsing image ' +
+      url.substring(url.lastIndexOf('/') + 1) +
+      '.  Please wait …')
```
camarasu: If I'm not mistaken, this is what takes most of the time. We should probably inform the user about it with a message like "Please wait, this may take a while..."

Portal/vip-visualization/src/main/java/fr/insalyon/creatis/vip/visualization/client/view/AbstractViewTab.java
wassong-iphc: This message is replaced by the next message "Loading and parsing image …".
This latter message is then removed by code in the javascript file ami-viewer.js "window.hideMessage()" in case of success, or by "window.setWarningMessage('Error loading ' + url + '<br />' + error)", in case an error is detected.

The messages only stay if there an internal AMI error that does not bubble up to our javascript code.

Portal/vip-portal/src/main/webapp/js/ami-viewer.js
wassong-iphc: It depends on the format of the image.  mhd/raw images are parsed faster than nifti images, for example.  For mhd/raw files, the most time is spent fetching the files from the SE.

But I don't mind changing the message.

camarasu: Ok, then let's use "Please wait, this may take a while..." for both messages (download and parsing)

Portal/vip-portal/pom.xml
```diff
@@ -148,6 +162,14 @@ knowledge of the CeCILL-B license and that you accept its terms.
...
               </configuration>
             </plugin>
+
+            <plugin>
+              <artifactId>maven-war-plugin</artifactId>
+              <version>3.2.3</version>
+              <configuration>
+                <packagingExcludes>WEB-INF/lib/*-sources.jar</packagingExcludes>
+              </configuration>
+            </plugin>
```
axlbonnet: Is it really necessary ? I thought that `<scope>provided</scope>` would have make the source jar absent from the war.

wassong-iphc: Yes, tests have shown that this is necessary, to exclude source jars from the war.  And they are 2 sources jar from libraries used that are also excluded with this configuration.

PR #208, opened 2019-09-03
Use the connection pool everywhere

wassong-iphc:


Portal/vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/dao/mysql/PlatformConnection.java
```diff
@@ -59,154 +59,164 @@
...
     }
 
-    public Connection getConnection() {
+    // We throw a SQLException and not a DAOException because the client of this
+    // function must anyway catch the SQLException when closing the connection.
+    // This avoids him to have to catch 2 types of exceptions.
+    public Connection getConnection() throws SQLException {
+        Connection connection = dataSource.getConnection();
+        Connection actual = ((javax.sql.PooledConnection)connection).getConnection();
+        logger.info("actual connection : " + actual.toString());
```
camarasu: I"m not sure to fully understand the aim of the "actual" Connection. Is it only for logging purposes or is it more (the doc says that "The connection pool manager calls this method when an application has called the method DataSource.getConnection and there are no PooledConnection objects available.") ?

Portal/vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/dao/mysql/PlatformConnection.java
```diff
@@ -59,154 +59,164 @@
...
+                createTable(connection,
+                            "VIPTermsOfuse",
+                            "idTermsOfuse INT(11) NOT NULL AUTO_INCREMENT, "
+                            + "date TIMESTAMP NULL, "
+                            + "text TEXT NULL, "
+                            + "PRIMARY KEY (idTermsOfuse)");
+
+                firstExecution = false;
+            } catch (SQLException e) {
+                logger.error("Error closing database connection", e);
```
camarasu: How can we be sure that the SQLException is only concerned with closing the connection? 
In case it is concerned with something else, is it always enough to catch and log it ?

Portal/vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/dao/mysql/PlatformConnection.java
```diff
@@ -59,154 +59,164 @@
...
 
-    private PlatformConnection() throws DAOException {
-
+    private PlatformConnection() {
         firstExecution = true;
         connect();
         createTables();
     }
 
     private void connect() {
```
camarasu: The connection is no longer retrieved in this method. Maybe we could rename it accordingly?

Portal/vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/dao/mysql/PlatformConnection.java
wassong-iphc: Yes, getting the actual connection was done only for logging purpose.  I'll remove the log, it's too verbose.

Portal/vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/dao/mysql/PlatformConnection.java
wassong-iphc: I am sure of it.  The only code change was to add the fetching of the connexion.  The exception was not needed before that, so the sole case were it could happen is when closing the connection.

The strategy used in the rest of this function is to catch and log exception, see what is done with the DAOException.  So I implemented the same strategy.  This code is used when the servlet is initialised.  I am not sure what can be done.  Throw a RuntimeException ?

Portal/vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/dao/mysql/PlatformConnection.java
wassong-iphc: What do you suggest to rename: the class name PlatformConnection, or the method name of connect() ?

PlatformConnection is still the singleton allowing the server to get a database connection.
connect() could be renamed to initDatasource().

camarasu: The method name. I find your suggestion, initDatasource(), very good, go for it.

Portal/vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/dao/mysql/PlatformConnection.java
camarasu: If the only possible exception concerns the connection closing, then I would say that catching and logging it should be fine. But what if getConnection fails to retrieve the connection, isn't that possible ?

wassong-iphc: You are right : the SQLException could also happen when getConnection is called.  I can change the message to : "Error while getting the connection or closing it".  Is that enough ?

camarasu: OK, go for the more explicit error message. I have no better suggestion.

PR #224, opened 2020-01-22
improve initial state of database after creation

axlbonnet:
Correct database creation requests and add objects that are necessary at first startup.


Portal/vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/dao/MySQLDAOFactory.java
```diff
@@ -104,6 +105,7 @@ private MySQLDAOFactory() {
                 connection,
                 "VIPApplications",
                 "name VARCHAR(255), "
+                    + "lfn varchar(255) DEFAULT NULL, "
```
camarasu: lfn in VIPApplications is obsolete, we'd better remove it. It has been replaced by lfn in VIPAppVersions

camarasu:
Great, thanks !

PR #228, opened 2020-01-28
Added checkBosh method (use of CVMFS when possible)

camarasu:
PR related to GASW PR#15

### DIFF
--- a/Portal/vip-portal/src/main/resources/vm/wrapper.vm
+++ b/Portal/vip-portal/src/main/resources/vm/wrapper.vm
@@ -112,17 +112,23 @@ cat << JSONPARAMETERS  > input_param_file.json
 {
 JSONPARAMETERS
 
+count=0
 while [[ $# > 0 ]]
 do
 key="$1"
 case $key in
     #foreach($input in $tool.getInputs())
 --$input.getId().toLowerCase())
-       #if($tool.hasNextInput($input))
-            echo "\"$input.getId()\" : \"$2\"," >> input_param_file.json
-       #else
-            echo "\"$input.getId()\" : \"$2\"" >> input_param_file.json
-       #end
+        if [ "$2" != "No_value_provided" ]
+        then
+            if [ "$count" -gt "0" ]
+            then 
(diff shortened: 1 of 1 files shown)


Portal/vip-portal/src/main/resources/vm/wrapper.vm
```diff
@@ -73,6 +73,35 @@ fi
...
+    info "CVMFS bosh in ${BOSH_CVMFS_PATH} not working, checking for a local version"
+    bosh create foo.sh 
+    if [ $? != 0 ]
+    then
+        info "bosh is not found in PATH or it is does not work fine, searching for another local version"
+        local HOMEBOSH=`find $HOME -name bosh`
+        if [ -z "$HOMEBOSH" ]
+        then
+            info "bosh not found, trying to install it"
+            pip install --trusted-host pypi.org --trusted-host pypi.python.org --trusted-host files.pythonhosted.org boutiques --prefix $PWD || error "pip install boutiques failed"
```
axlbonnet: In case of error (if there is not any working bosh), I think it will still try to execute it and fail ?

Portal/vip-portal/src/main/resources/vm/wrapper.vm
```diff
@@ -113,12 +148,9 @@ JSONPARAMETERS
 TOOLNAME="$tool.getName()"
 JSONFILE="${TOOLNAME}.json"
 
-#check bosh and install it if needed
-bosh create foo.sh || pip install --trusted-host pypi.org --trusted-host pypi.python.org --trusted-host files.pythonhosted.org boutiques --prefix $PWD || error "pip install boutiques failed"
-#we need to export PYTHONPATH and PATH because of the prefix installation
-export PYTHONPATH=$PYTHONPATH:$PWD/lib/python2.7/site-packages
```
axlbonnet: PYTHONPATH export has been removed. It is not necessary anymore ?

camarasu: Yes, I don't think it is useful

Issue #191, opened 2019-05-20
Migrate from Axis 1 to Axis2


fg-cdd:
Migrate from old Axis 1 library to new Axis2.

Module vip-application uses the library Axis 1, version 1.4.  This library is obsolete and has been replaced by Axis2, which is a rewrite with a new architecture.  The packages and the API have changed.  Migrating needs to change the code.  A migration guide is available here: https://axis.apache.org/axis2/java/core/docs/migration.html

The latest version of Axis2 is 1.7.9, dated nov. 2018.

PR #232, opened 2020-03-19
upgrade axis to a patched version

axlbonnet:
relative to #191 

It will take to much time to upgrade from axis1 to axis2 as it is a major rewrite that requires major changes in VIP too.

Actually, although there are several vulnerabilities in axis 1.4 (2006) and no release has been done since, patches are published (https://github.com/apache/axis1-java), build (https://travis-ci.org/github/apache/axis1-java) and archived (https://repository.apache.org/content/repositories/snapshots/org/apache/axis/axis/1.4.1-SNAPSHOT).

VIP now relies on these versions, so one need to install one of these patched version labelled as version "axis:1.4.1-vip" with one of these solution :
- install in a local maven repository
- install in a local nexus repository

### DIFF
--- a/Portal/vip-application/pom.xml
+++ b/Portal/vip-application/pom.xml
@@ -97,7 +97,7 @@ knowledge of the CeCILL-B license and that you accept its terms.
         <dependency>
             <groupId>org.apache.axis</groupId>
             <artifactId>axis</artifactId>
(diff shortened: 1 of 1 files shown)


camarasu:
Great, thank you!

PR #233, opened 2020-03-22
Cleaning

axlbonnet:
Clean mainly maven config :
- remove unnecessary configuration
- move some dependencies version configuration in parent pom
- remove code concerning shiwapool and the soap api (unused, not maintained) and their maven dependencies
- correct or ignore almost all java compiler warnings
- use maven provided scope to not include in final war gwt dependencies that are only needed at compile time (gwt-dev or smart-gwt)
- remove unnecessary leading Portal folder

All this allow to reduce the final war size from 165M to 80M


camarasu:
Great work !

PR #234, opened 2020-03-30
Replace log4j

axlbonnet:
This replaces log4j1.2 (end-of-life in 2015) by slf4j + logback :
- slf4j as facade, easy to switch the logging framework in the future
- easy to log parameters with slf4j
- same configuration and format for vip.log as before (except stacktraces shorter to improve readability)
- additional vip-errors.logs that duplicates error logs with more information
- logback allows for runtime changes to, for instance, silence a class that is too verbose or debug on a failing class

I also improved how exceptions are handled to log **all** exceptions in a coherent manner (all exceptions are logged once, no more, no less) with the following scheme :
- checked exceptions are logged at the lowest level with as much context as possible and transformed into VIP exceptions (CoreException, BusinessException, APIException, etc)
- VIP exceptions bubble up (DaoException into BusinessExcepton into CoreException) and are not logged at the highest server level
- RuntimeException's are catched and logged at the highest server lever (in REST controllers, in the GWT RPC parent servlet and in custom VIP servlets)


Portal/vip-api/src/main/java/fr/insalyon/creatis/vip/api/rest/controller/ApiKeyController.java
```diff
@@ -85,7 +85,6 @@ public ApiKeyController(
         try(Connection connection = connectionSupplier.get()) {
             return apiKeyBusiness.apiKeysFor(userEmail, connection);
         } catch (BusinessException | SQLException | SQLRuntimeException e) {
```
camarasu: If it's an SQLException, it's not logged anywhere else, is it ?

Portal/vip-visualization/src/main/java/fr/insalyon/creatis/vip/visualization/server/business/VisualizationBusiness.java
```diff
@@ -221,8 +227,7 @@ public VisualizationItem getVisualizationItemFromLFN(
                     extension));
         } catch (DataManagerException dme) {
             logger.warn(
```
camarasu: Do we want to log it here ?

camarasu:
Great job !

Issue #229, opened 2020-02-12
Make Dirac tag configurable for application importer


camarasu:
Make the list of tags belong to a config file instead of hard-coded

camarasu:
PR #235

Issue #226, opened 2020-01-24
Include version in the path of imported applications


axlbonnet:
When an application is imported from a boutiques file, it generated several files that are stored in the DFC. Currently the root directory of these files is chosen by the administrator that does the import, and  then the importer add another directory (with the application name), and some subdirectories (bin, workflow, gasw, json).

This is not adapted to applications with several versions as the administrator must create before a directory with the version (without dot as it is forbidden in the VIP portal). Also, as it is not easy to know where the previous versions were saved, it happens that different versions of the same applications are stored in very different places.

The improvements that must be done are the following :
- the importer also include the version in the imported application path, in the format {ApplicationName}/{ApplicationVersion}/[bin|workflow||gasw|json]
- the importer proposes a root directory to store the application by default, so that all application are stored in the same place. The LFN from this default location is the VIP configuration file. Although it isn't advised, it is possible to select another root directory.

PR #235, opened 2020-04-08
Configurable tags for app import (issues #226 and #229)

camarasu:
Improvements in application importer allowing to:
- configure a list of possible application requirements (i.e., dirac tags) in the VIP config file instead of hardcoding them;
- configure a default root path for the import of applications, so that they are all in the same place by default. 

issues #226 and #229

### DIFF
--- a/vip-application-importer/src/main/java/fr/insalyon/creatis/vip/applicationimporter/client/view/applicationdisplay/VIPLayout.java
+++ b/vip-application-importer/src/main/java/fr/insalyon/creatis/vip/applicationimporter/client/view/applicationdisplay/VIPLayout.java
@@ -30,30 +30,20 @@
  * knowledge of the CeCILL-B license and that you accept its terms.
  */
 package fr.insalyon.creatis.vip.applicationimporter.client.view.applicationdisplay;
-import com.google.gwt.json.client.JSONObject;
-import com.google.gwt.json.client.JSONParser;
 import com.google.gwt.user.client.rpc.AsyncCallback;
 import com.smartgwt.client.types.Alignment;
 import com.smartgwt.client.types.Overflow;
-import com.smartgwt.client.util.SC;
-import com.smartgwt.client.widgets.form.DynamicForm;
 import com.smartgwt.client.widgets.form.fields.CheckboxItem;
 import com.smartgwt.client.widgets.form.fields.SelectItem;
 import com.smartgwt.client.widgets.form.fields.TextItem;
 import com.smartgwt.client.widgets.form.fields.events.ChangeEvent;
-import fr.insalyon.creatis.vip.application.client.ApplicationConstants;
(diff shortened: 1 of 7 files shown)


vip-application-importer/src/main/java/fr/insalyon/creatis/vip/applicationimporter/client/view/applicationdisplay/VIPLayout.java
```diff
@@ -161,7 +176,29 @@ private SelectItem createTagsSelect() {
...
+                tagsCb.setValueMap("None");
+            }
+
+            @Override
+            public void onSuccess(List<String> result) {
+                if(!result.contains("None")){
+                    result.add("None");
+                }
+                
+                LinkedHashMap<String, String> requirementsValues = new LinkedHashMap<String, String>();
```
axlbonnet, line 192: Better to use the Map type for the declaration. 
And the diamond <> on construction should work. 
So I suggest
`Map<String, String> requirementsValues = new LinkedHashMap<>();`

vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/business/Server.java
```diff
@@ -245,6 +247,9 @@ private Server() {
             //Applicatoin importer
             applicationImporterFileRepository = config.getString(CoreConstants.APPLICATION_FILES_REPOSITORY, "/tmp/boutiques-cache");
             deleteFilesAfterUpload = config.getString(CoreConstants.APP_DELETE_FILES_AFTER_UPLOAD, "yes");
+            applicationImporterRootFolder = config.getString(CoreConstants.APP_IMPORTER_ROOT_FOLDER, "/biomed/user/c/creatis/vip/data/groups/Applications");
+            List<String> listApplicationImporterRequirements = new ArrayList<>();
+            applicationImporterRequirements = config.getList(CoreConstants.APP_REQUIREMENTS, listApplicationImporterRequirements);
```
axlbonnet, line 252: why not put `new ArrayList<>()` as `getList` 2nd parameter instead of using an useless variable ?

axlbonnet:
2 small remarks, fine to me otherwise !

vip-application-importer/src/main/java/fr/insalyon/creatis/vip/applicationimporter/client/view/applicationdisplay/VIPLayout.java
camarasu: Thank you for the suggestion. Done

vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/business/Server.java
camarasu: Done also.

Issue #204, opened 2019-07-08
API : restrict special character in execution parameters as in the portal


axlbonnet:

camarasu:
We should be able to accept inputs such as "-a [Source_ID,138][particle,e-][energy,0.24000][nb,100000000][seed,6384646]". Check this with Gilles M.

PR #238, opened 2020-04-24
Restrict characters in api

axlbonnet:
issues #203 and #204 

This changes :
- filename have the same treatments in the api as in the vip portal (spaces replaced by _, accents and non-ascii chars removed)
- execution inputs have the same treatments in the api as in the vip portal (only a short white list of characters) + the `[]` brackets as needed in some gate api executions

Also :
- uri are now decoded in the rest data api (percent encoding. %20 => ' '). this was a bug
- cleaning and accents removing is improved and more strict, as I see reason to remove accents but to keep non-ascii characters. And this should correct some issues with some users that had strange characters in their name resulting in bug on signup when creating their home folder #203 

Exemple of filename to test in the api :
- `max _ - aà aÀ aầ -« _» eé eè eêfin t~ 6œ cç`
- <==>
- `max%20_%20-%20a%C3%A0%20a%C3%80%20a%E1%BA%A7%20-%C2%AB%20_%C2%BB%20e%C3%A9%20e%C3%A8%20e%C3%AAfin%20t%7E%206%C5%93%20c%C3%A7`

### DIFF
--- a/vip-api/src/main/java/fr/insalyon/creatis/vip/api/CarminProperties.java
+++ b/vip-api/src/main/java/fr/insalyon/creatis/vip/api/CarminProperties.java
@@ -66,4 +66,5 @@ public interface CarminProperties {
     // CARMIN PROCESSING
 
     String API_PIPELINE_WHITE_LIST = "carmin.processing.pipelines.whitelist";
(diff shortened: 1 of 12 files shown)


camarasu:
Looks great to me!

PR #249, opened 2020-06-09
Improve api

axlbonnet:
- Update spring and others libraries versions.
- Simplify/refactor vip-api given that the soap stuff is removed
- Correct and add some integration tests


camarasu:
Looks great !

Issue #255, opened 2020-09-07
Correct purging rights


axlbonnet:
Only administrators must have the right to purge executions :
- verify it in the client on the home page timeline and in the execution monitor
- ensure it on server side

https://trello.com/c/5JrNMBWN

PR #256, opened 2020-09-08
Correct purge rights

axlbonnet:
Deal with issue #255


camarasu:
Looks good to me, thank you for fixing it.

Issue #257, opened 2020-09-15
Correct session handling


axlbonnet:
There are often errors linked to user sessions in client notification (`User not logged in`) or in server logs (`No VIP user found in session`)

There are 2 reasons, coming from client requests scheduled to run every n seconds (with timers) :
- on logout, several of these timers may not be canceled, and the requests fail on server because there is no user in the java session.
- on server restart, java session are erased and next requests will fail because the server cannot find the user in session. This concern clients that keep the vip portal open in the browser (no F5, no browser restart).

PR #258, opened 2020-09-16
Correct session management

axlbonnet:
Deal with #257 by :
- restoring session on server from the email/session-id stored in cookies (correct server restart issues)
- stop all client timers on logout


camarasu:
Look good to me, thanks!

PR #267, opened 2020-12-02
Use spring globaly

axlbonnet:
This is a major change in VIP architecture, introducing Spring-framework globally (it was already there in vip-api). The main change is to use spring to manage the creation and injection of dependencies, in a IoC (Inversion of Control) way and with dependency injection. Spring allows to do that with few changes, in a "convention over configuration" way and it is not invasive in the code base, relying mostly in annotations.
Concerning VIP, this concerns these changes :
- the "root" configuration class is `SpringCoreConfig` (configured in `web.xml`), and all the "magic" resides in the `@ComponentScan(basePackages = "fr.insalyon.creatis.vip")` line. When vip starts, Spring to scan all the vip classes and detect those annotated with spring annotations like `@Service`, `@Repository`, `@Component`. It will create an unique instance of these classes (called "beans", that have singleton comportment) and manage them its application context. It will also analyse these classes to detect there dependencies (in constructor or in `@Autowired` methods) and inject these instances in one another. So there is almost no more `new` keyword left in the code.
- there are some spring configuration classes (annotated with `@Configuration`) declaring beans coming from external librairies (grida, sma, workflowsdb)
- static singleton or static classes are replaced by spring-managed beans (not static but with a singleton comportment)
- few classes cannot comport as singleton because they hold an internal state and are not thread-safe. The have a special spring scope, "prototype" meaning spring creates a new instance at each call. This is like doing a "new", except that spring still manage these short-lived beans, injecting their dependencies and handling other services (mainly database-related, see bellow). These beans must be injected in a special way, with `@Lookup` methods, see details in the code
- the servlet classes (mainly the GWT servlets `xxxxxxServiceImpl`) are special because they are declared in the `web.xml` file, they are created by the servlet container (tomcat) and cannot be created by spring. The solution is to find the spring application context in the `init` method of these servlets to obtain the spring-managed dependencies to inject (see `AbstractRemoteServiceServlet`)

The other main change is how database connections are handled. Before, connections were created at the highest level and were transmitted as parameters all the way down until the DAO classes. Spring is able to do that automatically, no more `Connection` parameters boilerplate, and it does much more : clean connection opening/closing, and automatic transaction management. This works this way :
- we declare the database to use in a `Datasource` bean (looking by default in the jndi context as before). 
- we annotate with `@Transactional` the classes where the transactions begin (and end)
- we use spring utility in the dao to obtain the current connection
- And that's it. When a method of a `@Transactional` classes is called (typically a `xxxxBusiness` class), spring will begin a transaction and create a unique connection linked to that transaction. The transaction/connection will be used in all the dao methods callled inside that method and will live until the method ends, when the transaction will commit (or rollback if there is an exception).
This is the cause for much of the changes of this PR, which consist of removing all the unnecessary `Connection` parameters and all the connection creation and closing (with cumbersome try/catch). This results in a much clearer code, and more stable.

Spring allows for more minor but interesting changes :
- Spring handle the properties coming from property file and put them in its `Environment` bean, or make them injectable through `@Value` annotation. This makes it simpler to obtain these property value.
- Database creation (if needed) is done automatically on startup

Finally, as spring handles all the beans in its application context, it is now very easy to change some beans at will to control the vip comportment. Spring profiles help to do that. This allows for easy automatic testing :
- there are several `IT` junit tests, using spring to bootstrap vip with some necessary changes to remove all external constraint : in particular it uses a in-memory h2 database and it replaces grida/sma/myProxy/moteur clients by configurable mocks. I've made several tests to verify the database/transaction/connection behavior under spring management, but there is a base test class `BaseSpringIT` that just need to be extended to do automatic functional testing.
- there is a "vip-local" module, overriding services that need external services by basic local ones : this concerns mainly grida and moteur. "local" Grida stores files in a local repository, and "local" moteur read gwendia files and launches a bash script in local. This is basic, this could be much improved if needed, but this already allow to launch a local vip instance, totally independent from the production vip.


camarasu:
Excellent, great job !
As discussed, I merge the PR as such and we'll have another one concerning the updates for the config file.

Issue #268, opened 2020-12-03
Correct boutiques import/executions bugs


axlbonnet:
Includes :
- container (docker) information is not shown anymore on the application import file (old boutiques file evolution not taken into account)
- space in an application name (in the boutiques file) causes an import error, because the application generated files (gwedia/gasw/wrapper) have the space in the the path.
- boutiques types Number and Flag (boolean) in an application causes execution errors because parameters of these types are generated as strings (for instance `"numberParam":"42"`) in the boutiques parameter file (instead of json native types number and boolean - without the quotes, for instance `"numberParam":42`)
- optional parameters are wrongly taken into account on a relaunch (and on examples/saved inputs). They have the `No_value_provided` value that is considered as a real value in the application launch form.

PR #269, opened 2020-12-07
268 boutiques corrections

axlbonnet:
Based on the #267 PR that should be accepted first for a better readibility.

Several correction for #268. Some implementation notes :
- boutiques file allows several container technology, only docker is supported on VIP import
- space in a application name causes an importation error
- boutiques type "Number" are generated as json number in the json parameter file
- boutiques type "Flag" are generated as json boolean in the json parameter file, **but only** if it is true. It is absent if it is false, as this is imposed in the boutiques version currently in use. As Flag are optional, they can be absent by default, they can be present by default (with a `default-value:true` in the boutiques file), and this default behavior can be overridden by true/false value in the execution launch form. Note that a `default-value:false` in the boutiques file is valid but has no use and can cause a bug in a vip execution.
- `No_value_provided` in relaunch is not easy to take into account as there are many many cases possible and it seems not useful to make a big effort to cover everything. I choose a simple solution that covers the most important use cases : `No_value_provided` are ignored in relaunched execution or in examples (and do not override or activate the input) ; and any real value (other than `No_value_provided`) overrides and activates a deactivated input (without conflict).

Also includes a session recovery correction because there was a bug when a cookie was duplicated.


camarasu:
Looks good to me, thanks !

PR #276, opened 2021-01-19
not an error if no somme row in gatelab db

axlbonnet:
If there is no data in the `Somme` table, it is not considered an error anymore.

@camarasu But I am not sure what to do when there is an error (h2 server issue, no table because of workflow issue) : do we still silence it for the user and return 0 (as before), or do we throw an error (as proposed).


camarasu:
OK with throwing an error, it should catch our attention.

PR #277, opened 2021-01-28
improve local instance configuration

axlbonnet:
This makes simpler to launch a local vip instance. Now almost everything are easy or automatic : 
- valid config files are provided
- necessary folders are created automatically
- test classes automatically validate the configuration and creates a minimal working environment (engine / class / application)

Much is still to be done but this is improving.


camarasu:
It's really great to have a working local instance configuration, well done !

I think we can have further improvements, especially on the way applications are handled and executed, but this can be done independently from this PR.

Issue #299, opened 2021-09-06
Save inputs not working


draguar:
When trying to save inputs, I am getting :
Unable to verify execution name
BuisinessException:DAOException:SQLDataException: Current position is after the last row

alexcornier:
- Save inputs is working now: ✔️

axlbonnet:
Corrected in #347

PR #342, opened 2022-05-16
Save inputs

alexcornier:
Github issue: Save inputs not working #299


alexcornier:
This PR is closed because it contains other dev and will be replaces by another one.

Issue #254, opened 2020-09-02
Make VIP-Application fields compulsory


camarasu:
Fields marked with a * should be compulsory, but they aren't

camarasu:
This concerns the Publications module

alexcornier:
- Make the fields for the creation of a new publication mandatory (Title, Type, Typename, Authors, date, VIP Application used): ✔️

axlbonnet:
Done in #353

PR #345, opened 2022-06-09
publication mandatory

alexcornier:
Github issue Make VIP-Application fields compulsory #254


alexcornier:
This PR is closed because it contains other dev and will be replaces by another one.

PR #350, opened 2022-06-13
Extending External platfrom table with keycloak attributes

alaeessaki:
Extending External platfrom table with keycloak attributes used in scripts


vip-datamanagement/src/main/java/fr/insalyon/creatis/vip/datamanager/client/bean/ExternalPlatform.java
```diff
@@ -46,6 +46,9 @@ public enum Type {
     private Type type;
     private String description;
     private String url;
+    private String upload_url;
+    private String keycloak_client_id;
+    private String refresh_token_url;
```
axlbonnet, line 51: Please use camel case for fields names : `uploadUrl`

axlbonnet (changes requested):

vip-datamanagement/src/main/java/fr/insalyon/creatis/vip/datamanager/client/bean/ExternalPlatform.java
alaeessaki: Done, you can check the changes.

axlbonnet (approved):

PR #371, opened 2022-07-18
adding executions summary endpoint

alaeessaki:


vip-api/src/main/java/fr/insalyon/creatis/vip/api/KeycloakSpringSecurityConfig.java
```diff
@@ -122,6 +122,7 @@ protected void configure(HttpSecurity http) throws Exception {
             .authorizeRequests()
                 .antMatchers("/rest/platform").permitAll()
                 .antMatchers("/rest/authenticate").permitAll()
+                .antMatchers("/rest/executions/{executionId}/summary").permitAll() // We can secure it by IP, but it's changed because of reverse proxy
```
axlbonnet, line 125: Why not authorize the shanoir IP like for the `/rest/register` URL ?

alaeessaki: the IP of the shanoir microservice machine is either hidden or changed because of Nginx reverse proxy as described in the description and the comment, 
I tested the IP whitelisting. when I do the curl within the machine (with the whitelisted IP) it works, but behind an Nginx reverse proxy, it doesn't.  

I think we can fix this if we call VIPs URL `http://vip-url/rest/executions/summary` directly from the shanoir microservice instand of the reversed proxy URL `https://shanoir-ng-front/shanoir-ng/vip/rest/executions/summary`
But is this a good thing to do?

What do you think?

axlbonnet: At first glance, I don't see any objection to call VIP from the internal URL, as it should stay in a secure network.
I prefer that to a public URL giving access to some possibly sensitive data.
I wanted to explore the keycloak features to have a token-secured communication between Shanoir and VIP but have not found the time to do it.

axlbonnet:
To repeat my comment, I cannot accept this PR in this state, it is not possible for an API endpoint giving access to personal data to be public.
I suggest that this endpoint be protected by IP first and we will accept this PR
Then we will find a way to protect the 2 endpoints dedicated to Shanoir by keycloak.

alaeessaki:
Hello Axel, 
I agree with your suggestion, we secure it with the IP, and find a solution for the reverse proxy.
You can check the new commit.
with kind regards.

PR #372, opened 2022-08-25
Make selectable text in the msg error & increase the time

alexcornier:
Make the text selectable in the error message and increase the message appearance time to 20 seconds
GI: Improve error message printing time #363

### DIFF
--- a/vip-core/src/main/java/fr/insalyon/creatis/vip/core/client/view/common/MessageWindow.java
+++ b/vip-core/src/main/java/fr/insalyon/creatis/vip/core/client/view/common/MessageWindow.java
@@ -69,10 +69,10 @@ public class MessageWindow {
         panel.addDrawHandler(new DrawHandler() {
             @Override
             public void onDraw(DrawEvent event) {
+                messageLabel.setCanSelectText(true);
                 panel.setWidth(messageLabel.getWidth());
                 panel.setHeight(messageLabel.getHeight());
                 panel.moveTo(getPanelXPosition(canvas), 2);
-                messageLabel.setCanSelectText(true);
             }
         });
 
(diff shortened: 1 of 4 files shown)


vip-core/src/main/java/fr/insalyon/creatis/vip/core/client/view/common/MessageWindow.java
```diff
@@ -123,6 +124,7 @@ private void createLabel(String title, String icon) {
             public void onDraw(DrawEvent event) {
                 messageLabel.setWidth(messageLabel.getVisibleWidth());
                 messageLabel.setMargin(5);
+                messageLabel.setCanSelectText(true);
```
axlbonnet, line 127: This is called 3 times in this class. Only 1 must be enough.
I think only this call should be kept, on it would be better to call it above after `setContents`

vip-core/src/main/java/fr/insalyon/creatis/vip/core/client/view/user/account/PasswordLayout.java
```diff
@@ -100,7 +100,7 @@ public void onClick(ClickEvent event) {
                                 @Override
                                 public void onFailure(Throwable caught) {
                                     WidgetUtil.resetIButton(saveButton, "Save Changes", CoreConstants.ICON_SAVED);
-                                    Layout.getInstance().setWarningMessage("Unable to update password:<br />" + caught.getMessage());
```
axlbonnet, line 103: keep this one unlimited as this is an unexpected error.

vip-core/src/main/java/fr/insalyon/creatis/vip/core/client/view/user/account/PasswordLayout.java
```diff
@@ -100,7 +100,7 @@ public void onClick(ClickEvent event) {
...
+                                    Layout.getInstance().setWarningMessage("Unable to update password:<br />" + caught.getMessage(), 10);
```
axlbonnet, line 103: keep this one unlimited as this is an unexpected error.

axlbonnet (changes requested):
The error when the password is wrong in `SigninTab` should also be temporary at 10 seconds

axlbonnet (approved):

PR #375, opened 2022-09-14
Udocker version quick fix

camarasu:


vip-portal/src/main/resources/vm/wrapper.vm
```diff
@@ -38,9 +39,7 @@ function download_udocker {
   cat >docker <<'EOF'
         #!/bin/bash
         MYARGS=$*
-        echo "my args are: " $MYARGS
-        #MYARGS=${MYARGS/"--entrypoint=/bin/sh"}
-        #echo "my latest args are: " $MYARGS
+        info "executing ./udocker/udocker/udocker $MYARGS"
```
axlbonnet: Our own command `info` is not available/exported here en this context and this executes the bash command `info`.
We should use classic `echo`

PR #389, opened 2022-10-13
get institution with check in using the scope "eduperson_scoped_affiliation"

alexcornier:
When a new user logs in via check-in, the insitution is directly filled in using the eduperson_scoped_affiliation scope. If this is equal to "null" then the field is filled with "Unknow". The user keeps the possibility to fill in the institution field afterwards

### DIFF
--- a/vip-api/src/main/java/fr/insalyon/creatis/vip/api/controller/EgiController.java
+++ b/vip-api/src/main/java/fr/insalyon/creatis/vip/api/controller/EgiController.java
@@ -51,8 +51,19 @@ public class EgiController {
         }
 
         Map<String, Object> userAttributes = authToken.getPrincipal().getAttributes();
+        Object institution = userAttributes.get("eduperson_scoped_affiliation");
+        String domainName;
+        if(institution != null){
+            String institution_string = institution.toString();
+            String temp = institution_string .substring(institution_string .indexOf("@") + 1);
+            domainName = temp.substring(0, temp.indexOf("."));
+            domainName = domainName.toUpperCase();
+        } else {
+            domainName = "Unknown";
+        }
+
         try {
-            User vipUser = configurationBusiness.getOrCreateUser((String) userAttributes.get("email"), userAttributes.get("eduperson_scoped_affiliation"), null);
+            User vipUser = configurationBusiness.getOrCreateUser((String) userAttributes.get("email"), domainName, null);
             SecurityContextHolder.clearContext(); // destroy spring session and use VIP own session mechanism
             vipSessionBusiness.setVIPSession(request, response, vipUser); // creates VIP cookies and session
         } catch (BusinessException | CoreException e) {
--- a/vip-api/src/main/java/fr/insalyon/creatis/vip/api/controller/EgiController.java
+++ b/vip-api/src/main/java/fr/insalyon/creatis/vip/api/controller/EgiController.java
@@ -52,7 +52,7 @@ public class EgiController {
 
         Map<String, Object> userAttributes = authToken.getPrincipal().getAttributes();
         try {
-            User vipUser = configurationBusiness.getOrCreateUser((String) userAttributes.get("email"), null);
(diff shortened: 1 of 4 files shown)


vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/business/ConfigurationBusiness.java
```diff
@@ -386,7 +386,7 @@ private User signin(String email, String password, boolean resetSession)
         }
     }
 
-    public User getNewUser(String email, String firstName, String lastName) {
+    public User getNewUser(String email, String firstName, String lastName, String egi_affiliation) {
```
axlbonnet, line 389: We are in internal VIP business class that must not have link with egi
Just name the argument/parameter `institution` everywhere in this class

vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/business/ConfigurationBusiness.java
```diff
@@ -985,15 +986,24 @@ public User getOrCreateUser(String email, String defaultAccountType)
                 }
             }
 
-            user = getNewUser(email, firstName, lastName);
+            if(egi_affiliation != null){
```
axlbonnet, line 989: Put this business into EgiController as this is specific to egi

axlbonnet (changes requested):

axlbonnet:
We also need to do some testing with google-like check-in account and non-cnrs people.

vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/business/ConfigurationBusiness.java
```diff
@@ -985,15 +986,24 @@ public User getOrCreateUser(String email, String defaultAccountType)
                 }
             }
 
-            user = getNewUser(email, firstName, lastName);
+            if(institution != null){
```
axlbonnet, line 989: Move this block to `EgiController`, in a new method called before `getOrCreateUser`.

This should not be there, because `getOrCreateUser` must receive a good institution in a string (like `CNRS`) but not a email or any strange format.

axlbonnet (changes requested):

vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/business/ConfigurationBusiness.java
alexcornier: Done

vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/business/ConfigurationBusiness.java
alexcornier: Done

vip-api/src/main/java/fr/insalyon/creatis/vip/api/controller/EgiController.java
```diff
@@ -51,8 +51,19 @@ public RedirectView getOauth2LoginInfo(
         }
 
         Map<String, Object> userAttributes = authToken.getPrincipal().getAttributes();
+        Object institution = userAttributes.get("eduperson_scoped_affiliation");
+        String domainName;
+        if(institution != null){
+            String institution_string = institution.toString();
```
axlbonnet, line 57: no `_`, put `institutionString`

vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/auth/AbstractAuthenticationService.java
```diff
@@ -147,7 +147,7 @@ private void authSuccessResponse(
             String email) throws BusinessException, CoreException {
 
         String accountType = getDefaultAccountType();
-        User user = configurationBusiness.getOrCreateUser(email, accountType);
+        User user = configurationBusiness.getOrCreateUser(email,null,  accountType);
```
axlbonnet, line 150: Won't it generate a NullPointerException later, maybe put "Unknown" ?

PR #390, opened 2022-10-24
Make mandatory Select type of Application

alexcornier:
Import application - no type of application #295

### DIFF
--- a/vip-application-importer/src/main/java/fr/insalyon/creatis/vip/applicationimporter/client/view/applicationdisplay/DisplayTab.java
+++ b/vip-application-importer/src/main/java/fr/insalyon/creatis/vip/applicationimporter/client/view/applicationdisplay/DisplayTab.java
@@ -48,6 +48,9 @@ import fr.insalyon.creatis.vip.applicationimporter.client.view.Constants;
 import fr.insalyon.creatis.vip.core.client.view.ModalWindow;
 import fr.insalyon.creatis.vip.core.client.view.layout.Layout;
 import fr.insalyon.creatis.vip.core.client.view.util.WidgetUtil;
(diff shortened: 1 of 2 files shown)


vip-application-importer/src/main/java/fr/insalyon/creatis/vip/applicationimporter/client/view/applicationdisplay/VIPLayout.java
```diff
@@ -136,7 +135,11 @@ public boolean getIsRunOnGrid() {
      * @return the type
      */
     public String getApplicationType() {
-        return appCbItem._getValue().toString();
+        if (appCbItem._getValue() == null){
+            return "Unknown";
```
axlbonnet, line 139: It is better to return `null` than to create a new constant that will be copy/parted

vip-application-importer/src/main/java/fr/insalyon/creatis/vip/applicationimporter/client/view/applicationdisplay/DisplayTab.java
```diff
@@ -106,7 +106,11 @@ private void configure() {
             @Override
             public void onClick(ClickEvent event) {
                 boutiquesTool.setApplicationLFN(vipLayout.getApplicationLocation() + "/" + boutiquesTool.getName());
-                createApplication();
+                if (vipLayout.getApplicationType() == "Unknown"){
```
axlbonnet, line 109: It is better to do the verification first and to `return` if it fails.

axlbonnet (changes requested):

axlbonnet (approved):

PR #394, opened 2023-01-06
Adapt security to be compatible with service account

alaeessaki:
The "SERVICE" role can be temporary for the moment. we can specify a role in the future


axlbonnet:
Fine to me.
If the `ShanoirProperties.SHANOIR_HOST_IP` property is not used anymore anywhere, could you also delete it ?

alaeessaki:
Dear Axel, 
Thank you. for your review, you can check the new changes.
Regards.

PR #410, opened 2023-03-16
modifications on the privacy policy & terms of use

alexcornier:


camarasu:
Thank you @alexcornier !

Issue #413, opened 2023-04-07
Bugfix : private group listed in public application 


axlbonnet:
If a app is linked to a public group AND to a private group (through classes), then both groups are displayed but the private group must not.

PR #414, opened 2023-04-20
small correction for the list of public classes

alexcornier:
Bugfix : private group listed in public application #413


axlbonnet (approved):
Nice lambda !

PR #415, opened 2023-05-11
Strip extension

SandeshPatilCreatis:
1. Boutiques support stripping the extension of the file. ('path-template-stripped-extensions')
2. But this feature of the boutique schema is not functioning on VIP.
3. It is to remove the extension of the path template (similar to baseline command in bash).

1. descriptor -> xml

    Descriptor key : Mapped to Java class (BoutiquesOutputFile) 2 variables , one for capturing strip extn values and another boolean variable which decide the descriptor contains strip extn or not and defined getter setters for those two variables
    To set value of those 2 variables in BoutiquesOutputFile : Java class : BoutiquesParser method : parseBoutiquesOutputFile and in gasw.vm : to add strip extn value in template tag based on condition

2. xml -> file creation with /without strip extn

    GaswParser : to read strip extn value from xml. if it’s not present in xml then create empty object of Set<String>
    GaswOutputTemplatePart : created a varaible to store value to srip extensions(since there can be multiple extensions we are using Set<String> stripExtensions)
    GaswParser : wherever we create object of GaswOutputTemplatePart class, we pass strip extension value to the constructor.
    GaswParser: method parseOutputTemplat: Inside case : Name , we replace strpextension with blank.(removing strip extensions from i/p file)


vip-portal/src/main/resources/vm/gasw.vm
```diff
@@ -31,9 +31,15 @@
 #set ($value=$value.replace($input.getValueKey(),"$na$i"))
 #end
 #end
-      <template value="$prefix1$dir1/$na1/$value$options1" />
-      <access type="LFN"/>
-    </output>
-#end
-  </executable>
+#set ($i=$i+1)
```
axlbonnet, line 34: I think this line should not be here

vip-portal/src/main/resources/vm/gasw.vm
```diff
@@ -31,9 +31,15 @@
...
+
+#if($outputFile.isStripExtension())
+    <template value="$prefix1$dir1/$na1/$value$options1" strip-extension="$outputFile.getPathTemplateStrippedExtensions()"/>
```
axlbonnet, line 37: Could you add the `if` to only add the strip-extension part ? Like

```
     <template value="$prefix1$dir1/$na1/$value$options1"
     #if($outputFile.isStripExtension())
          strip-extension="$outputFile.getPathTemplateStrippedExtensions()"
     #end 
    />
```

vip-datamanagement/src/main/java/fr/insalyon/creatis/vip/datamanager/client/bean/ExternalPlatform.java
```diff
@@ -40,7 +40,8 @@ public class ExternalPlatform {
 
     public enum Type {
         GIRDER,
-        SHANOIR
+        SHANOIR,
+        SRM
```
axlbonnet, line 44: SRM stuff should not be in this PR

vip-application/src/main/java/fr/insalyon/creatis/vip/application/client/bean/boutiquesTools/BoutiquesOutputFile.java
```diff
@@ -47,7 +49,10 @@ public class BoutiquesOutputFile implements IsSerializable {
     private boolean list;
     private boolean optional;
     private String commandLineFlag;
-
+    private Set<String> pathTemplateStrippedExtensions;
+    private boolean isStripExtension;
```
axlbonnet, line 53: This boolean is not necessary.
I prefer you to use a condition on `pathTemplateStrippedExtensions` being not null and not empty in the gasw.vm instead.

axlbonnet (changes requested):
This is working but I suggest some small improvements.

Also, there are 2 PR mixed in there, the SRM stuff should not be present here. Could you do a new PR, starting from the current develop and containing only  the stripped extensions stuff ?

SandeshPatilCreatis:
redundant #421

PR #423, opened 2023-07-03
gfal_support (SrmStorageBusiness - USERS_HOME to lower case)

SandeshPatilCreatis:
changed DataManagerConstants.USERS_HOME to lowerCase for gfal path with srm protocol.


SandeshPatilCreatis:
duplicate

PR #439, opened 2023-07-26
correct VIP api tests

MathildeMrz:


MathildeMrz:
This should not include github actions files.

PR #449, opened 2023-09-01
Cvmfs

sandepat:
Some CVMFS containers tries create temporary files inside the container, which is prevented by cvmfs due to read-only access to the storage elements, resulting in failure of the execution (e.g BratsPipeline)

1. Make a directory named tmp in $PWD and bind mount it on the container to avoid any creation of temporary files in case of a cvmfs container.
2. For the containers pulled from docker-hub (or similar repositories), should also use the $PWD/tmp folder

### DIFF
--- a/vip-portal/src/main/resources/vm/wrapper.vm
+++ b/vip-portal/src/main/resources/vm/wrapper.vm
@@ -161,7 +161,16 @@ fi
 echo "import sys; sys.setdefaultencoding(\"UTF8\")" > sitecustomize.py
 #creating a temporary directory for bindmount /tmp of containers
 mkdir -p tmp
-PYTHONPATH=".:$PYTHONPATH" $BOSHEXEC exec launch $JSONFILE input_param_file.json -v $PWD/../cache:$PWD/../cache -v $PWD/tmp:/tmp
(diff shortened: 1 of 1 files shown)


axlbonnet:
After some thoughts, there are some issues with this solution.
Firstly, boutiques/bosh automatically bind $PWD when launching, so $PWD/tmp would be bound at 2 different paths and this may cause issues.
More importantly, a docker container will likely create some file as root in the "/tmp" folder, and moteur/gasw cannot clean/delete this files afterwards (this is specific to local executions).

With Sorina, we have thought about an improved solution :
- the TMP_FOLDER path will be different depending if the pipeline executes on the grid or on local resources
- the TMP_FOLDER name will be specific to the job, to avoid parallel job conflicts : `TMP_FOLDER=tmp_$(basename $PWD)`
- in local, the TMP_FOLDER will be in /tmp : `TMP_FOLDER=/tmp/$TMP_FOLDER`
- on the grid, the TMP_FOLDER will be in the parent directory (the job "root" directory) : `TMP_FOLDER=../$TMP_FOLDER`
- the TMP_FOLDER will be mounted as `/tmp` on the container
- the TMP_FOLDER will be deleted after the container execution (ignore errrors)

Issue #119, opened 2017-04-07
You cannot view 'Support' group


vincent-legoll:
In the 'Groups' tab, if you double click on the 'Support group', you get the message 'You cannot edit Support group', but the detailed view is not updated, for example to view that goup's details.

That message looks msplaced in the group workflow, it should only be thrown at the user if he actually try tu make modifications, or when he click the save button. This would allow to view the group's details.

axlbonnet:
I propose :
- make the 'Support group' visible
- disable the 'save/remove' button for it.

axlbonnet:
Also verify if this group is useful and could be replaced.

axlbonnet:
implemented in #470 :
- it is meant to be a list of users who receive support mail, nothing else
- support group is editable by admins (but must remain private)
- support group cannot be associated to a class

To discuss with @camarasu

PR #470, opened 2024-03-12
Manage Support group properly

axlbonnet:
see #119


axlbonnet:
TODO : check if the support group is necessary and if it used differently from the admin list.

axlbonnet:
Not needed anymore.

PR #474, opened 2024-05-16
Added support for Singularity

sandepat:
1. Updated the wrapper script to link singularity to CVMFS apptainer image (to execute singularity commands)
2. CVMF apptainer image - /cvmfs/dirac.egi.eu/container/apptainer/bin/apptainer

### DIFF
--- a/vip-application/src/main/java/fr/insalyon/creatis/vip/application/client/view/boutiquesParsing/BoutiquesParser.java
+++ b/vip-application/src/main/java/fr/insalyon/creatis/vip/application/client/view/boutiquesParsing/BoutiquesParser.java
@@ -1,12 +1,19 @@
 package fr.insalyon.creatis.vip.application.client.view.boutiquesParsing;
 
+import java.util.Map;
+import java.util.Set;
+
 import com.google.gwt.json.client.JSONArray;
 import com.google.gwt.json.client.JSONObject;
 import com.google.gwt.json.client.JSONParser;
-import fr.insalyon.creatis.vip.application.client.bean.boutiquesTools.*;
-import java.util.HashSet;
-import java.util.Map;
-import java.util.Set;
+
+import fr.insalyon.creatis.vip.application.client.bean.boutiquesTools.BoutiquesApplication;
+import fr.insalyon.creatis.vip.application.client.bean.boutiquesTools.BoutiquesFlagInput;
+import fr.insalyon.creatis.vip.application.client.bean.boutiquesTools.BoutiquesGroup;
+import fr.insalyon.creatis.vip.application.client.bean.boutiquesTools.BoutiquesInput;
+import fr.insalyon.creatis.vip.application.client.bean.boutiquesTools.BoutiquesNumberInput;
+import fr.insalyon.creatis.vip.application.client.bean.boutiquesTools.BoutiquesOutputFile;
+import fr.insalyon.creatis.vip.application.client.bean.boutiquesTools.BoutiquesStringInput;
 
 /**
  * Helper class for parsing JSON objects
@@ -102,6 +109,15 @@ public class BoutiquesParser extends AbstractJsonParser{
             } catch (InvalidBoutiquesDescriptorException exception){
                 throw new InvalidBoutiquesDescriptorException("Invalid Boutiques descriptor: invalid container-image.",
                         exception);
+            }
+        }
+        // Custom property
+        JSONObject customObject = getObjectValue(parsedDescriptor, "custom", true);
+        if (customObject != null) {
+            try {
+                application.setVipContainer(getStringValue(customObject, "vip:container", true));
+            } catch (InvalidBoutiquesDescriptorException exception) {
+                throw new InvalidBoutiquesDescriptorException("Invalid Boutiques descriptor: invalid vip:container.");
             }
         }
         // Json descriptor
@@ -228,4 +244,6 @@ public class BoutiquesParser extends AbstractJsonParser{
         bof.setCommandLineFlag(commandLineFlag);
         return bof;
     }
+
+
 }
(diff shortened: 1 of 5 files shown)


vip-portal/src/main/resources/vm/wrapper.vm
```diff
@@ -19,6 +19,23 @@ function error {
...
+function function_singularity {
+    cat > singularity << 'EOF'
+          #!/bin/bash
+          /cvmfs/dirac.egi.eu/container/apptainer/bin/apptainer "$@"
+EOF
+}
+
+# Call the function to create the singularity script and make it executable
+function_singularity
+chmod +x singularity
```
axlbonnet, line 31: as for udocker, better do a `chmod a+x`.
also, move this line inside the `function_singularity` method to change the rights just after you create it

vip-portal/src/main/resources/vm/wrapper.vm
```diff
@@ -19,6 +19,23 @@ function error {
   echo [ ERROR - $D ] $* >&2
 }
 
+function function_singularity {
+    cat > singularity << 'EOF'
+          #!/bin/bash
+          /cvmfs/dirac.egi.eu/container/apptainer/bin/apptainer "$@"
```
axlbonnet, line 25: Not a good idea to have an hardcoded path like that. We will use a environment variable for the apptainer location, like `APPTAINER_LOCATION`.

vip-portal/src/main/resources/vm/wrapper.vm
```diff
@@ -19,6 +19,23 @@ function error {
...
+EOF
+}
+
+# Call the function to create the singularity script and make it executable
+function_singularity
+chmod +x singularity
+export APPTAINER_PULLFOLDER="/cvmfs/dirac.egi.eu/container/apptainer/"
```
axlbonnet, line 32: Is it the env variable that apptainer uses to find images ?

vip-portal/src/main/resources/vm/wrapper.vm
```diff
@@ -19,6 +19,23 @@ function error {
...
+
+# Define a function to run singularity commands
+singularity() {
+    ./singularity "$@"
+}
```
axlbonnet, line 37: It is better to add `PWD` to the `$PATH` like it's done for udocker

vip-portal/src/main/resources/vm/wrapper.vm
```diff
@@ -19,6 +19,23 @@ function error {
...
 
+function function_singularity {
+    cat > singularity << 'EOF'
+          #!/bin/bash
+          /cvmfs/dirac.egi.eu/container/apptainer/bin/apptainer "$@"
+EOF
+}
+
+# Call the function to create the singularity script and make it executable
+function_singularity
```
axlbonnet, line 30: We have the choice to always add singularity/apptainer (like you did), or to do it only when the boutiques app uses singularity. (like it's done for udocker).
@camarasu what do you prefer ?

vip-portal/src/main/resources/vm/wrapper.vm
sandepat: Updated to chmod a+x singularity

vip-portal/src/main/resources/vm/wrapper.vm
sandepat: yes. Should this also be added to the configuration file?

vip-portal/src/main/resources/vm/wrapper.vm
sandepat: Done. The path for the $APPTAINER_LOCATION should be declared in gaswConfiguration?

vip-portal/src/main/resources/vm/wrapper.vm
sandepat: done.

vip-portal/src/main/resources/vm/wrapper.vm
sandepat: done

vip-application/src/main/java/fr/insalyon/creatis/vip/application/client/view/boutiquesParsing/BoutiquesParser.java
```diff
@@ -104,6 +111,15 @@ public BoutiquesApplication parseApplication(String descriptor) throws InvalidBo
                         exception);
             }
         }
+        // Custom property
+        JSONObject customObject = getObjectValue(parsedDescriptor, "custom", true);
+        if (customObject != null) {
+            try {
+                application.setVipContainer(getStringValue(customObject, "vip:imagepath", true));
+            } catch (InvalidBoutiquesDescriptorException exception) {
```
axlbonnet, line 119: i think we can avoid this exception catching, as the key is optional and the new exception does not add useful information to the original one.

vip-portal/src/main/resources/vm/wrapper.vm
```diff
@@ -187,7 +203,12 @@ TMP_FOLDER=/tmp/$TMP_FOLDER
 TMP_FOLDER=../$TMP_FOLDER
 #end
 mkdir -p $TMP_FOLDER
+
+#if($imagepath == "")
```
axlbonnet, line 207: does it work if `imagepath` is null ?

axlbonnet (commented):
almost ready !
Just a minor improvement and a interogation.

vip-portal/src/main/resources/vm/wrapper.vm
sandepat: If $imagepath is null or undefined in, the #if condition will handle it as if it were an empty string (""). 
In VTL, a null value and an empty string are treated similarly in conditions. So I think the statement handles both conditions appropriately.

vip-application/src/main/java/fr/insalyon/creatis/vip/application/client/view/boutiquesParsing/BoutiquesParser.java
sandepat: Done

camarasu:
Well done, thank you!

PR #486, opened 2024-08-29
add property myproxy.enabled

ethaaalpha:
Adding the property inside the vip.conf to avoid the `Error reading proxy file /vip/.vip/proxies/biomed/x509up_server` when not using dirac.

### DIFF
--- a/vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/business/simulation/WebServiceEngine.java
+++ b/vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/business/simulation/WebServiceEngine.java
@@ -104,7 +104,7 @@ public class WebServiceEngine extends WorkflowEngineInstantiator {
             throws
             java.rmi.RemoteException,
             javax.xml.rpc.ServiceException {
-		String strProxy = null;
+        String strProxy = null;
 
         System.setProperty("javax.net.ssl.trustStore", server.getTruststoreFile());
         System.setProperty("javax.net.ssl.trustStorePassword", server.getTruststorePass());
@@ -112,9 +112,9 @@ public class WebServiceEngine extends WorkflowEngineInstantiator {
 
         // String settings = "This is going to contain settings...";
         // Get Proxy from current User's context
-		if (server.getMyProxyEnabled()) {
-			strProxy = ProxyUtil.readAsString(proxyFileName);
-		}
+        if (server.getMyProxyEnabled()) {
+            strProxy = ProxyUtil.readAsString(proxyFileName);
+        }
 
         //configure WS with the specific wsdd file (fix pb of conflict with grid service)
         String resourcename = "moteur-client-config.wsdd";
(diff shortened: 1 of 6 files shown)


vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/business/simulation/WebServiceEngine.java
```diff
@@ -104,14 +104,17 @@ public String launch(String proxyFileName, String userDN)
             throws
             java.rmi.RemoteException,
             javax.xml.rpc.ServiceException {
+		String strProxy = null;
```
axlbonnet, line 107: seems you used tab indentation.
we use 4-spaces indent in vip.
Optionnaly you can try to convince me to change to tab but you need arguments.

axlbonnet: I think you initialize the var here as a good practice to declare all local variable at the beginning of the function ?

axlbonnet: Have you verified that `null` is supported and does not trigger an error in the vip/moteur communication ?

vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/business/proxy/ProxyClient.java
```diff
@@ -95,7 +95,8 @@ public ProxyClient(Server server) {
     public void checkProxy() throws BusinessException {
 
         int proxyHoursValidity = server.getMyProxyMinHours();
-        if (proxyHoursValidity == 0) {
+
+        if (!server.getMyProxyEnabled()) {
```
axlbonnet, line 99: I like to add spaces around `!` to make it more visible if that's fine to you.

vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/business/SpringConfigServer.java
```diff
@@ -244,6 +244,11 @@ public int getMyProxyMinHours() {
         return env.getRequiredProperty(CoreConstants.LAB_MYPROXY_MIN_HOURS, Integer.class);
     }
 
+	@Override
+	public boolean getMyProxyEnabled() {
+		return env.getRequiredProperty(CoreConstants.LAB_MYPROXY_ENABLED, Boolean.class);
```
axlbonnet, line 249: I had in mind to make it optional (but true by default).
And also to make the other `myproxy.*` config required only if `myproxy.enabled ` is true (or absent)

axlbonnet (changes requested):
Very good work.
See the few details in comments :
- indentation style
- optional config if possible.

vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/business/simulation/WebServiceEngine.java
ethaaalpha: > I think you initialize the var here as a good practice to declare all local variable at the beginning of the function ?

Yes

ethaaalpha: > Have you verified that `null` is supported and does not trigger an error in the vip/moteur communication ?

I've run few jobs on the vip portal and it all works well, the vip-errors.log and moteur_server.log were empty.

axlbonnet (approved):

PR #509, opened 2024-12-04
Update application packaging doc.

ngeorges-cnrs:
Replace import_application.html content to point to the newly created github wiki page.

PR #510, opened 2024-12-05
Update application packaging doc.

ngeorges-cnrs:
Replace import_application.html content to point to the newly created github wiki page.

This PR replaces https://github.com/virtual-imaging-platform/VIP-portal/pull/509, which had the right diff but an incorrect source branch.


camarasu (approved):
Looks good to me, thank you!

PR #512, opened 2024-12-06
Support for multiple OIDC/Keycloak servers

ngeorges-cnrs:
This patch is a follow-up of the #503 PR, which solves issue #489.
The inner structure to support multiple servers was already there in #503, so the changes are limited to three things :
- Replacement of the "keycloak.json" config file and its single-server definiton with a new file "vip-oidc.json". I've removed keycloak.json from README.md, the new file name and its format should probably documented somewhere, but README.md feels too public for such technical detail, so for now the new file format is documented in a comment section in OidcConfig.java : to be discussed.
Examples :
```
# old file : .vip/keycloak.json 
{
  "auth-server-url":"https://some_keycloak_server",
  "realm":"test"
}
# equivalent new file : .vip/vip-oidc.json
{
  "servers":[{
    "url":"https://some_keycloak_server/realms/test"
  }]
}
```

- Addition of very basic tests for JWT authentication, based on Spring mocks. These were completed by a semi-manual test process which uses curl + real Keycloak servers, which could be documented in the same technical doc as the config file format.

- And finally one extra change, independent from the multi-servers issue, but related to java21/SpringSecurity changes : by adding a new log in `ApiSecurityConfig.CurrentUserProvider`, the #503 patch caused a verbose log message any time VIP-portal gets an unauthenticated request to a public API endpoint, such as `GET /rest/platform` :
```
2024-11-29 04:53:52,189 |ERROR |http-nio-8080-exec-10 | f.i.c.v.a.s.ApiSecurityConfig$CurrentUserProvider:170 - CurrentUserProvider: unknown principal class class java.lang.String
```
I tracked this down to the fact that SpringSecurity enables [anonymous authentication](https://docs.spring.io/spring-security/reference/servlet/authentication/anonymous.html) by default, so on an unauthenticated request, principal is in fact not null but is a String instead. This was likely already the case in previous java11 version, where we just didn't log anything.
I decided to solve this by disabling anonymous authentication (commit https://github.com/virtual-imaging-platform/VIP-portal/commit/f4a23c3d75c9848811bf48cd09ef75d6d936bf32), so that principal is now null on unauthenticated requests, and unexpected principal instances continue to be logged. Alternative valid solutions would be to either remove the logging entirely (i.e. back to the java11 implementation), or to whitelist the anonymous auth with something such as `principal instanceof String && (String) principal == "anonymousUser"`. But as far as I can tell this anonymous mode serves no purpose in our usage, which is why I just disabled it.


ngeorges-cnrs:
No code change on this PR, which remains theoretically mergeable as is, but a complement : I've updated VIP-portal wiki with a new page related to Keycloak/OIDC API authentication : https://github.com/virtual-imaging-platform/VIP-portal/wiki/API-authentication.
This wiki page is meant to replace the README.md section that this PR removes. It provides the reference documentation for the `vip-oidc.json` config file, as well as some resources to manually deploy and test a VIP + Keycloak setup with bearer token authentication.

Besides validating this doc, two related open questions are :
- Removing some of the `vip-oidc.json` info in the big comment block in OidcConfig.java, now redundant with the wiki
- Removing the `vip-api/src/test/resources/vip-keycloak.postman_collection.json` file, which I think was only referenced by the old README section.

I'll wait for feedback before updating the PR.

vip-api/src/test/java/fr/insalyon/creatis/vip/api/rest/itest/SpringAuthenticationIT.java
```diff
@@ -111,4 +99,20 @@ public void authenticationWithoutCredentials() throws Exception {
                 .andExpect(jsonPath("$.errorCode")
                         .value(ApiError.INSUFFICIENT_AUTH.getCode()));
     }
+
+    @Test
+    public void authenticationWithInvalidBearerToken() throws Exception {
+        mockMvc.perform(get("/rest/wrongUrl")
+                        .with(new BearerTokenRequestPostProcessor("invalidToken")))
+                .andExpect(status().isUnauthorized())
+                .andExpect(content().contentType(MediaType.APPLICATION_JSON_VALUE));
```
axlbonnet, line 108: Why not verifying the VIP error code as in the previous tests ?

vip-api/src/test/java/fr/insalyon/creatis/vip/api/rest/itest/SpringAuthenticationIT.java
```diff
@@ -111,4 +99,20 @@ public void authenticationWithoutCredentials() throws Exception {
...
+    }
+
+    @Test
+    public void authenticationWithValidBearerToken() throws Exception {
+        mockMvc.perform(get("/rest/wrongUrl")
+                        .with(SecurityMockMvcRequestPostProcessors.jwt()))
```
axlbonnet, line 114: This generates a valid JWT token for spring ?
I guess this works in the case because we don't need a VIP user ?

axlbonnet (commented):
Just a few remarks about the code.

I agree to not duplicate documentation and to keep it only on the wiki. So:
- the comment about the json format can me removed
- the old `vip-keycloak.postman_collection.json`  was not used and can be removed

vip-api/src/main/java/fr/insalyon/creatis/vip/api/security/oidc/OidcConfig.java
```diff
@@ -53,38 +53,63 @@ public OidcConfig(Environment env, Resource vipConfigFolder) throws IOException,
...
+                String resourceName;
+                if (serverNode.hasNonNull("resource")) {
+                    resourceName = serverNode.get("resource").asText();
+                } else {
+                    resourceName = "";
+                }
+                // Add a new OIDC server to our config.
+                URI url = new URI(baseURL);
+                String issuer = url.toASCIIString();
+                if (servers.get(issuer) != null) {
```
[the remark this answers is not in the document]
ngeorges-cnrs: Done

vip-api/src/main/java/fr/insalyon/creatis/vip/api/security/oidc/OidcConfig.java
```diff
@@ -53,38 +53,63 @@ public OidcConfig(Environment env, Resource vipConfigFolder) throws IOException,
...
+                if (serverNode.hasNonNull("use-resource-role-mapping")) {
+                    useResourceRoleMappings = serverNode.get("use-resource-role-mapping").asBoolean();
+                } else {
+                    useResourceRoleMappings = false;
+                }
+                String resourceName;
+                if (serverNode.hasNonNull("resource")) {
+                    resourceName = serverNode.get("resource").asText();
+                } else {
+                    resourceName = "";
```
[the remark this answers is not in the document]
ngeorges-cnrs: Done

vip-api/src/test/java/fr/insalyon/creatis/vip/api/rest/itest/SpringAuthenticationIT.java
ngeorges-cnrs: Unintentional omission, fixed.

vip-api/src/test/java/fr/insalyon/creatis/vip/api/rest/itest/SpringAuthenticationIT.java
ngeorges-cnrs: If I understand https://docs.spring.io/spring-security/site/docs/current/api/org/springframework/security/test/web/servlet/request/SecurityMockMvcRequestPostProcessors.html#jwt() correctly, it doesn't even really generate a JWT, it just sets a principal and accepts it.
This test is clearly not worth much more than nothing at all, but doing anything significant with JWTs seems non-trivial, as actual JWT generation and verification needs private/public keys, and we'd also need to mock the OidcConfig object.

So going further into this seemed to be lots of efforts to end up with mocks anyways, which would still not check that the "real thing" works. I've tried to document relatively precisely how to test an actual setup in https://github.com/virtual-imaging-platform/VIP-portal/wiki/API-authentication#keycloak-configuration. It's a fairly simple process, the longest part is configuring Keycloak, and this can be automated if we want to.

axlbonnet: agreed

PR #517, opened 2025-01-15
Add singularityPath

ngeorges-cnrs:
See https://github.com/virtual-imaging-platform/GASW/pull/88

### DIFF
--- a/vip-portal/src/main/resources/vm/wrapper.vm
+++ b/vip-portal/src/main/resources/vm/wrapper.vm
@@ -173,6 +173,8 @@ fi
 # Check that singularity is in PATH
 
 #if($tool.getContainerType() == "singularity")
+# command -v displays the path to singularity if found
+info "checking for singularity"
 if ! command -v singularity
 then
     if test -d "$singularityPath"; then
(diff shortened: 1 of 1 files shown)


vip-portal/src/main/resources/vm/wrapper.vm
```diff
@@ -180,11 +170,17 @@ then
...
 #if($tool.getContainerType() == "singularity")
 if ! command -v singularity
 then
-    configure_apptainer_for_singularity
+    if test -d "$singularityPath"; then
+        export PATH="$singularityPath:$PATH"
+        info "adding singularityPath to PATH ($singularityPath)"
+    else
+        warning "singularityPath not found ($singularityPath), leaving PATH unchanged"
+    fi
```
axlbonnet, line 183: Could we just add a debug to display the absolute path of the singularity command if it is exists ?

ngeorges-cnrs: https://github.com/virtual-imaging-platform/VIP-portal/pull/517/commits/98f7274f9223e2e367d25aedf38d07c4b990f502: `command -v` already display the path if found, but it was indeed a bit implicit. I've made that explicit in a comment, and added a log line above for context.

PR #522, opened 2025-02-18
admin of group able to watch apps executions

ethaaalpha:
From now on the portal, admin group users can watch executions related to the applications associated (through a class object naturally)

### DIFF
--- a/vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/business/WorkflowBusiness.java
+++ b/vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/business/WorkflowBusiness.java
@@ -381,7 +381,7 @@ public class WorkflowBusiness {
         List<String> applications = (application != null) ? Collections.singletonList(application) : new ArrayList<>();
 
         List<Simulation> simulations = new ArrayList<>();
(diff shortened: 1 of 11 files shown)


vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/business/WorkflowBusiness.java
```diff
@@ -394,7 +394,10 @@ public List<Simulation> getSimulationsWithGroupAdminRights(User user, String app
             workflows.addAll(getSimulationsAdminGroup(user.getEmail(), applications, wStatus, appClass, startDate, endDate, tag));
             workflows.addAll(workflowDAO.get(users, applications, wStatus, appClass, startDate, endDate, tag));
 
-            simulations = parseWorkflows(new ArrayList<>(workflows));
+            workflows = new ArrayList<>(workflows.stream().collect(
+                Collectors.toMap(Workflow::toString, w -> w, (e, r) -> e)).values());
```
axlbonnet, line 398: Maybe add a small comment to explain why you do that.
Also why not using the workflow-id as key ?

PR #524, opened 2025-02-20
Data API fixes, using Grida getPathInfo

ngeorges-cnrs:
This patch uses the new `getPathInfo()` Grida command (https://github.com/virtual-imaging-platform/GRIDA/pull/44) to fix some existing bugs in VIP Data API regarding file/directory testing. In practice, visible changes for a VIP-API client are as follows :

On `GET /rest/path?action=list` :
- the API call was failing on a directory that contained a single item named as its parent. It now works.
- the API call was reporting `HTTP 200` with a body of `[]` when listing a non-existing path. It now correctly fails with `HTTP 400`.
- these fixes come at the cost of an extra `COM_GET_PATH_INFO` before the existing `COM_LIST_FILES_AND_FOLDERS`

On `GET /rest/path?action=properties` :
- the `isDirectory` property was incorrectly `false` on a directory that contained a single item named as its parent. It is now `true`.
- all other situations should be unchanged, and the number of Grida calls is the same as before, since `COM_EXIST` was replaced with  `COM_GET_PATH_INFO`

### DIFF
--- a/vip-api/src/main/java/fr/insalyon/creatis/vip/api/business/DataApiBusiness.java
+++ b/vip-api/src/main/java/fr/insalyon/creatis/vip/api/business/DataApiBusiness.java
@@ -52,7 +52,6 @@ import fr.insalyon.creatis.vip.datamanager.server.business.LFCPermissionBusiness
 import fr.insalyon.creatis.vip.datamanager.server.business.TransferPoolBusiness;
 import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.io.input.ReaderInputStream;
-import org.apache.commons.io.input.ReaderInputStream.Builder;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -71,6 +70,7 @@ import java.util.ArrayList;
 import java.util.Base64;
 import java.util.List;
 import java.util.Locale;
+import java.util.Optional;
 import java.util.concurrent.*;
 import java.util.function.Supplier;
 
@@ -150,26 +150,28 @@ public class DataApiBusiness {
         }
         PathProperties pathProperties = new PathProperties();
         pathProperties.setPath(path);
-        Data.Type type = baseGetPathInfo(path);
-        if (type == Data.Type.folder) { // path is a directory
-            pathProperties.setExists(true);
-            List<Data> fileData = baseGetFileData(path);
-            pathProperties.setIsDirectory(true);
-            pathProperties.setSize((long) fileData.size());
-            pathProperties.setLastModificationDate(
-                    baseGetFileModificationDate(path) / 1000);
-            pathProperties.setMimeType(env.getProperty(CarminProperties.API_DIRECTORY_MIME_TYPE));
-        } else if(type == Data.Type.file) { // path is a regular file
-            pathProperties.setExists(true);
-            List<Data> fileData = baseGetFileData(path);
+        Optional<Data.Type> type = baseGetPathInfo(path);
+        if (type.isEmpty()) { // path doesn't exist
+            pathProperties.setExists(false);
+            return pathProperties;
+        }
+        pathProperties.setExists(true);
+        List<Data> fileData = baseGetFileData(path);
+        if (type.get() == Data.Type.file) {
+            // this is a file, not a directory
             Data fileInfo = fileData.get(0);
             pathProperties.setIsDirectory(false);
             pathProperties.setSize(fileInfo.getLength());
             pathProperties.setLastModificationDate(
                     getTimeStampFromGridaFormatDate(fileInfo.getModificationDate()));
             pathProperties.setMimeType(getMimeType(path));
-        } else { // path does not exist
-            pathProperties.setExists(false);
+        } else {
+            // its a directory
+            pathProperties.setIsDirectory(true);
+            pathProperties.setSize((long) fileData.size());
+            pathProperties.setLastModificationDate(
+                baseGetFileModificationDate(path) / 1000);
+            pathProperties.setMimeType(env.getProperty(CarminProperties.API_DIRECTORY_MIME_TYPE));
         }
         return pathProperties;
     }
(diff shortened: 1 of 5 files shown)


ngeorges-cnrs:
Extra commit https://github.com/virtual-imaging-platform/VIP-portal/pull/524/commits/1e9b4b71a66d48a7f94a9294f12fabc2f655cd45 to remove unused `uploadFileToSE` methods, which were causing a build error in VIP-portal `GridaClientLocal` override since the merge of https://github.com/virtual-imaging-platform/GRIDA/pull/45.

Also, CI should now be ok.

vip-api/src/main/java/fr/insalyon/creatis/vip/api/business/DataApiBusiness.java
```diff
@@ -150,13 +150,14 @@ public PathProperties getPathProperties(String path ) throws ApiException {
...
-        if (!baseDoesFileExist(path)) {
+        Optional<Data.Type> type = baseGetPathInfo(path);
+        if (type.isEmpty()) { // path doesn't exist
             pathProperties.setExists(false);
             return pathProperties;
         }
         pathProperties.setExists(true);
         List<Data> fileData = baseGetFileData(path);
-        if (doesPathCorrespondsToAFile(path, fileData)) {
+        if (type.get() == Data.Type.file) {
```
axlbonnet, line 160: better to avoid `==` and use `.equals(...)`

vip-api/src/main/java/fr/insalyon/creatis/vip/api/business/DataApiBusiness.java
```diff
@@ -180,11 +181,11 @@ public List<PathProperties> listDirectory(String path) throws ApiException {
         if (path.equals(ROOT)) {
             return getRootSubDirectoriesPathProps();
         }
-        List<Data> directoryData = baseGetFileData(path);
-        if (doesPathCorrespondsToAFile(path, directoryData)) {
+        if (!baseIsDirectory(path)) {
```
axlbonnet, line 184: why not re-using `type` to avoid a new grida call ?

axlbonnet: Actually is this condition even possible ?

vip-api/src/main/java/fr/insalyon/creatis/vip/api/business/DataApiBusiness.java
```diff
@@ -531,6 +522,19 @@ private boolean baseDoesFileExist(String path) throws ApiException {
         }
     }
 
+    private boolean baseIsDirectory(String path) throws ApiException {
+        Optional<Type> type = baseGetPathInfo(path);
+        return type.isPresent() && type.get() == Data.Type.folder;
```
axlbonnet, line 527: I think i prefer using the `Data.Type` type for more clarity.
Also, use `equals` instead of `==`

vip-api/src/main/java/fr/insalyon/creatis/vip/api/business/DataApiBusiness.java
```diff
@@ -271,12 +272,12 @@ private void checkDownloadPermission(String path) throws ApiException {
             logger.error("cannot download root ({})", path);
             throw new ApiException("Illegal data API access");
         }
-        List<Data> fileData = baseGetFileData(path);
-        if (!doesPathCorrespondsToAFile(path, fileData)) {
+        if (baseIsDirectory(path)) {
```
axlbonnet, line 275: Actually it doesn't verify if the path actually exist.
Maybe it is easier to call `baseGetPathInfo` and handle the 3 cases here.

axlbonnet (changes requested):
Just a few remarks.

vip-api/src/main/java/fr/insalyon/creatis/vip/api/business/DataApiBusiness.java
axlbonnet: Sorry, it was an error in my review.

vip-api/src/main/java/fr/insalyon/creatis/vip/api/business/DataApiBusiness.java
```diff
@@ -180,11 +181,11 @@ public List<PathProperties> listDirectory(String path) throws ApiException {
         if (path.equals(ROOT)) {
             return getRootSubDirectoriesPathProps();
         }
-        List<Data> directoryData = baseGetFileData(path);
-        if (doesPathCorrespondsToAFile(path, directoryData)) {
+        if (!baseIsDirectory(path)) {
             logger.error("Trying to list {} , but is a file :", path);
             throw new ApiException("Error listing a directory");
         }
+        List<Data> directoryData = baseGetFileData(path);
```
axlbonnet, line 188: Actually, in the error case just above, it can be a directory or a not-existing path ?
We should either add a condition for absent paths or update the error message.

vip-api/src/main/java/fr/insalyon/creatis/vip/api/business/DataApiBusiness.java
ngeorges-cnrs: Done in https://github.com/ngeorges-cnrs/VIP-portal/commit/7323ddee1596a5890418354d70db0512667f1275

vip-api/src/main/java/fr/insalyon/creatis/vip/api/business/DataApiBusiness.java
ngeorges-cnrs: Done in https://github.com/ngeorges-cnrs/VIP-portal/commit/7323ddee1596a5890418354d70db0512667f1275

vip-api/src/main/java/fr/insalyon/creatis/vip/api/business/DataApiBusiness.java
ngeorges-cnrs: Done https://github.com/ngeorges-cnrs/VIP-portal/commit/2e45b94de1e9502f67d97d1366cfd0d5ee5e5244.

PR #526, opened 2025-02-25
fix relative and absolute "file:" paths

ethaaalpha:
An issue occurred since we were adding "file:" to every local paths and it was breaking the realpath <-> basepath conversion.
The issue is fixed + units tests related.


vip-datamanagement/src/test/java/fr/insalyon/creatis/vip/datamanager/server/business/LfcPathsBusinessTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+    @Mock
+    private GroupDAO groupDAO;
+
+    @Mock
+    private User user;
+
+    private LfcPathsBusiness lfcPathsBusiness;
+
+    @BeforeEach
```
axlbonnet, line 32: Is it necessary before each test or only once ?

PR #531, opened 2025-03-26
fix monitor order

ethaaalpha:


vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/business/WorkflowBusiness.java
```diff
@@ -395,8 +395,9 @@ public List<Simulation> getSimulationsWithGroupAdminRights(User user, String app
             workflows.addAll(workflowDAO.get(users, applications, wStatus, appClass, startDate, endDate, tag));
 
             // this is to avoid duplicates
-            workflows = new ArrayList<>(workflows.stream().collect(
-                Collectors.toMap(Workflow::getId, w -> w, (e, r) -> e)).values());
+            workflows = new ArrayList<>(workflows.stream()
+                .sorted(Comparator.comparing(Workflow::getStartedTime).reversed())
+                .collect(Collectors.toMap(Workflow::getId, w -> w, (e, r) -> e)).values());
```
axlbonnet, line 400: Actually I'm not sure a map a ordered, so maybe converting to map can change order ?

ethaaalpha: I agree with you that maybe it don't work be cause of the map.
I'll suggest to add the ˋ.sort` inside the `parseWorkflows` method ? (since it's converted to list<simulation> inside the method)

PR #548, opened 2025-05-27
Add /rest/admin API

ngeorges-cnrs:
This PR :
- Fixes two regressions introduced in the `groups-update` branch:
  - Fix NPE in `AppVersionBusiness.getVersion()`, when an app version doesn't exist.
  - Fix failing DAO query in `ApplicationData.removeVersion()`.
- Adds two new API endpoints `/rest/admin/applications` and `/rest/admin/appVersions`, only for admin-level users. At this point, things are kept very simple, and these endpoints are just basic CRUD wrappers over existing business helpers :
  - This gives maximum flexibility to the only existing client, `vipapps.py` : https://github.com/virtual-imaging-platform/VIP-python-client/pull/46
  - Existing `/rest` API and business code is left unchanged to avoid regressions
  - `Application` and `AppVersion` beans are directly exposed with the following semantics :
    - `GET /rest/admin/applications` : get list
    - `GET /rest/admin/applications/<id>` : get single object
    - `POST /rest/admin/applications` or `PUT /rest/admin/applications/<id>` : create a new object if `id` doesn't exist, update it if it does. `id` is obtained from request body on `POST`. This ensures idempotency on both creation and update.
    - `DELETE /rest/admin/applications/<id>` : delete existing object. Success on unknown `id` ensures idempotency.

Many details and open questions should be addressed in a future update before this API can be considered for wider use, but none of them is critical for the current scope of `vipapps.py`. For instance : paging on GET, POST/PUT semantics, intermediate model objects vs direct beans exposure, ...

### DIFF
--- a/vip-api/src/main/java/fr/insalyon/creatis/vip/api/controller/admin/AppVersionController.java
+++ b/vip-api/src/main/java/fr/insalyon/creatis/vip/api/controller/admin/AppVersionController.java
@@ -90,20 +90,15 @@ public class AppVersionController extends ApiController {
     }
 
     private AppVersionStrings parseAppVersionId(String appVersionId) throws ApiException {
-        try {
-            int delimiterPos = appVersionId.lastIndexOf("/");
-            if (delimiterPos >= 0) {
-                String appName = appVersionId.substring(0, delimiterPos);
-                String version = appVersionId.substring(delimiterPos + 1);
-                return new AppVersionStrings(appName, version);
-            } else {
-                throw new ApiException("Invalid appVersionId");
-            }
-        } catch (ApiException e) {
-            logger.error("Error decoding appVersionId {}", appVersionId, e);
-            throw new ApiException("cannot decode appVersionId : " + appVersionId);
+        int delimiterPos = appVersionId.lastIndexOf("/");
+        if (delimiterPos >= 0) {
+            String appName = appVersionId.substring(0, delimiterPos);
+            String version = appVersionId.substring(delimiterPos + 1);
+            return new AppVersionStrings(appName, version);
+        } else {
+            logger.error("Error decoding appVersionId {}", appVersionId);
+            throw new ApiException(ApiException.ApiError.INVALID_PIPELINE_IDENTIFIER, appVersionId);
         }
-
     }
 
     @RequestMapping(value = "{appVersionId}", method = RequestMethod.GET)
@@ -113,7 +108,7 @@ public class AppVersionController extends ApiController {
         try {
             AppVersion appVersion = appVersionBusiness.getVersion(input.appName, input.version);
             if (appVersion == null) {
-                throw new ApiException("Not found");
+                throw new ApiException(ApiException.ApiError.INVALID_PIPELINE_IDENTIFIER, appVersionId);
             }
             return appVersion;
         } catch (BusinessException e) {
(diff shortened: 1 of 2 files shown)


vip-api/src/main/java/fr/insalyon/creatis/vip/api/controller/admin/AppVersionController.java
```diff
@@ -0,0 +1,176 @@
...
+            if (delimiterPos >= 0) {
+                String appName = appVersionId.substring(0, delimiterPos);
+                String version = appVersionId.substring(delimiterPos + 1);
+                return new AppVersionStrings(appName, version);
+            } else {
+                throw new ApiException("Invalid appVersionId");
+            }
+        } catch (ApiException e) {
+            logger.error("Error decoding appVersionId {}", appVersionId, e);
+            throw new ApiException("cannot decode appVersionId : " + appVersionId);
```
axlbonnet, line 104: is the exception raised in line 100 and catched just bellow or am i missing something ?

axlbonnet: You can reuse existing error codes : `throw new ApiException(ApiError.INVALID_PIPELINE_IDENTIFIER, identifier);`

vip-api/src/main/java/fr/insalyon/creatis/vip/api/controller/admin/AppVersionController.java
```diff
@@ -0,0 +1,176 @@
...
+    public AppVersion createOrUpdateAppVersion(@PathVariable String appVersionId,
+                                               @RequestBody @Valid AppVersion appVersion) throws ApiException {
+        logMethodInvocation(logger, "createOrUpdateAppVersion", appVersionId);
+        AppVersionStrings input = parseAppVersionId(appVersionId);
+        try {
+            AppVersion existingAppVersion = appVersionBusiness.getVersion(input.appName, input.version);
+            if (existingAppVersion == null) {
+                appVersionBusiness.add(appVersion);
+            } else {
+                appVersionBusiness.update(appVersion);
```
axlbonnet, line 140: don't we verify the name and version in the AppVersion object correspond to the one on the URL ?

vip-api/src/main/java/fr/insalyon/creatis/vip/api/controller/admin/ApplicationController.java
```diff
@@ -0,0 +1,118 @@
...
+    @RequestMapping(value = "/{applicationId}", method = RequestMethod.PUT)
+    public Application createOrUpdateApplication(@PathVariable String applicationId,
+                                                 @RequestBody @Valid Application app) throws ApiException {
+        logMethodInvocation(logger, "createOrUpdateApplication", applicationId);
+        try {
+            Application existingApp = applicationBusiness.getApplication(applicationId);
+            if (existingApp == null) {
+                applicationBusiness.add(app);
+            } else {
+                applicationBusiness.update(app);
```
axlbonnet, line 96: same here, do we verify the app name correspond ?

ngeorges-cnrs: Done.

PR #549, opened 2025-05-28
fix visibility match between app/resource and groups

ethaaalpha:
- added some logs also


camarasu:
User access rigths to applications and ressources are managed through groups, which can be private or public. Is there a reason for duplicating the private/public property at the application and ressource level?

PR #556, opened 2025-06-05
Add source field in AppVersions

ngeorges-cnrs:
This PR adds a new "source" field in AppVersions :
- this is a purely informative text field meant to document where the descriptor comes from.
- it is available for read/write in the web UI and the `/rest/admin` API. It is not interpreted for anything by VIP-portal.

Some points of attention :
- I used `TEXT` and not `VARCHAR(N)` as the SQL type, as URLs (a common value for this field) can be quite long.
- This PR also contains a very minor fix for the default value of the read-only descriptorField in EditVersionLayout.java, when adding a version manually from the web UI (an operation not done in practice)
- Impact on upcoming migrations :
  - schema change `ALTER TABLE VIPAppVersions ADD COLUMN source TEXT;` should be commented out in `migrate_db_*.sql` scripts whenever this PR is merged : see [here](https://github.com/virtual-imaging-platform/vip-scripts/blob/master/migration_v4/migrate_db_eskemm.sql#L56) and [there](https://github.com/virtual-imaging-platform/vip-scripts/blob/master/migration_v4/migrate_db_creatis.sql#L59)
  - a "source" column has been added to the CSV indexes that are planned for apps import. With a recent enough version of `vipapps.py sync_index`, this will set the source field to the content of that column. If the target instance of VIP-portal doesn't support the source field, then the column is just ignored.

### DIFF
--- a/vip-application/src/main/java/fr/insalyon/creatis/vip/application/client/bean/AppVersion.java
+++ b/vip-application/src/main/java/fr/insalyon/creatis/vip/application/client/bean/AppVersion.java
@@ -53,33 +53,35 @@ public class AppVersion implements IsSerializable {
     private List<String> resources;
     private List<String> tags;
     private Map<String, String> settings;
+    private String source;
 
     public AppVersion() {}
 
     public AppVersion(String applicationName, String version, String descriptor,
-                      Map<String,String> settings, boolean visible) {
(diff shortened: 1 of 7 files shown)


camarasu:
Looks good to me, thank you.

PR #562, opened 2025-06-25
Updated Gate inputs according to new Gate descriptor

camarasu:
First quick update attempt needs further improvements.
Aims at compatibility with https://github.com/virtual-imaging-platform/vip-apps-boutiques-descriptors/blob/main/GATE/GATE-9.4.1.json but does not set the jobNumber input, which is hidden and should have a default value of 1 to be overriden by VIP.


ngeorges-cnrs:
I tried running VIP-portal with this patch and it didn't work for me, with 4 issues encountered and fixed :
- the form just doesn't show (infinite spinner)
- then there is a JS error `ReferenceError: mainMacFileName is not defined` in `macroParser.js`
- then the `jobNumber` field is not initialized correctly
- then we lack the `results-directory` value, causing Moteurlite to fail

These fixes are visible here: https://github.com/ngeorges-cnrs/VIP-portal/commit/6dd61d737a324a0d891d5b2c370457b7eadb114f, but are FYI, and clearly deserves an extra look by someone more familiar with the topic.

Especially, there were two fixes with a possible functional impact :
- Moteurlite needs `results-directory` to be set, I did this with `BoutiquesApplicationExtensions(true)` but we may want to do this another way and adjust the directory value.
- `jobNumber` must be set to the same value as `numberOfJobs`, otherwise the number of iterations will be wrong. I did this by leaving the `jobNumber` field modifiable in the UI, thus pushing the responsibility on the user. This problem will disappear once `vip:overriddenInputs` is implemented, which is almost done on https://github.com/ngeorges-cnrs/VIP-portal/tree/overriddenInputs

camarasu:
Thank you @ngeorges-cnrs for the suggested fixes. Looks good to me, I've just integrated them.

PR #595, opened 2025-09-24
implement /internal/session endpoint

ethaaalpha:
Firstly, this pull request is based on another one: [multi_api](https://github.com/virtual-imaging-platform/VIP-portal/pull/594).
This one introduce the new `/internal/session` endpoint, which implement the DELETE, POST, GET methods.

### Spring Security
A new `SecurityFilterChain` is present:`InternalSecurityConfig`, it handles all /_internal/**_ urls:
- **CORS policy** is enabled
- **CSRF Token** verification is enabled: 
The token will be defined with *set_cookie* after a successful login. 
All of the "non secure" http requests will need the CSRF Token except for `POST on /internal/session`.
- `vip-session-cookie` and `vip-user-cookie` are defined on login and use by the filter chain to check authentication

### Cookies and actual frontend endpoints evolutions
By the way, we changed the cookies to `httpOnly` and adapt the actual frontend to support them in this case. It mean that the actual frontend will use `GET /internal/session` to determine if the user is actually logged (the current sign-in also use the new endpoint).
Some small modifications were made in *GWT serverside* code to support `httpOnly`.
The cookies are also set as secure if the `apache.sslport` properties is not *80*.

### Others
Some new annotations were introduced for the deprecation of *SmartGWT* (`@VIPRemoval` and `@VIPCheckRemoval`)

### DIFF
--- a/vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/controller/SessionController.java
+++ b/vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/controller/SessionController.java
@@ -14,6 +14,7 @@ import org.springframework.web.bind.annotation.RestController;
 import fr.insalyon.creatis.vip.core.client.bean.User;
 import fr.insalyon.creatis.vip.core.client.view.CoreConstants;
 import fr.insalyon.creatis.vip.core.server.business.BusinessException;
+import fr.insalyon.creatis.vip.core.server.business.ConfigurationBusiness;
 import fr.insalyon.creatis.vip.core.server.business.SessionBusiness;
 import fr.insalyon.creatis.vip.core.server.exception.ApiException;
 import fr.insalyon.creatis.vip.core.server.model.AuthenticationCredentials;
@@ -27,23 +28,28 @@ import jakarta.validation.Valid;
 public class SessionController {
 
     private final SessionBusiness sessionBusiness;
+    private final ConfigurationBusiness configurationBusiness;
     final private Supplier<User> userProvider;
 
     @Autowired
-    public SessionController(SessionBusiness sessionBusiness, Supplier<User> userProvider) {
+    public SessionController(SessionBusiness sessionBusiness, Supplier<User> userProvider, ConfigurationBusiness configurationBusiness) {
         this.sessionBusiness = sessionBusiness;
         this.userProvider = userProvider;
+        this.configurationBusiness = configurationBusiness;
     }
 
     @GetMapping
     public Session getSession(HttpServletRequest request, HttpServletResponse response) throws ApiException {
-        Session session = sessionBusiness.getSession(userProvider.get());
+        User user = userProvider.get();
+        Session session = sessionBusiness.getSession(user);
 
         try {
             // renew existing cookies
             sessionBusiness.createLoginCookies(request, response, session);
+            configurationBusiness.updateUserLastLogin(user.getEmail());
+
             return session;
-        } catch (UnsupportedEncodingException e) {
+        } catch (UnsupportedEncodingException | BusinessException e) {
             throw new ApiException("Failed to retrieve user session!", e);
         }
     }


vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/controller/SessionController.java
```diff
@@ -0,0 +1,92 @@
...
+import org.springframework.security.web.csrf.CookieCsrfTokenRepository;
+import org.springframework.security.web.csrf.CsrfToken;
+import org.springframework.web.bind.annotation.DeleteMapping;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+import fr.insalyon.creatis.vip.core.client.view.CoreConstants;
```
ngeorges-cnrs, line 16: New controllers for the internal API should ideally have no dependency on GWT-related packages, so any import of `*.client.*` or `*.server.rpc.*` should in theory induce a refactoring. This could also be handled a as separate task later.

vip-portal/src/main/webapp/js/home.js
```diff
@@ -1,25 +1,9 @@
...
-        return true;
-    } else {
-        window.location.href="index.html";
-        return false;
-    }
+function checkSession() {
+    fetch("/internal/session").then(function (response) {
+        if (response.status != 200) {
+            window.location = "index.html";
+        }
```
ngeorges-cnrs, line 5: Minor: maybe use `response.ok` instead of `response.status` ?

Also, some notes on JS `fetch()` error handling, which will be more important for the JS front-end :
- a 4XX/5XX response other than 401 technically doesn't indicate that the user is disconnected, so showing the login page in that case is not great, especially on 5XX. Some user-visible error message while staying on "home" might be preferable.
- there is a similar problem in case of network errors, where the promise won't resolve and which can be handled with `.catch()`

ngeorges-cnrs (commented):
Works for me in general, I just made minor comments on some code.

Two points to be discussed with @axlbonnet :
- the `/internal` API response body on errors is not final in the present PR, this is normally not a blocker, as it can be handled later.
- `server.getApacheSSLPort() != 80` as the condition for `httpOnly` cookies should probably be replaced by a dedicated new setting (and getApacheSSLPort removed as unused ?). This is used in two places where cookies are set, `VipSessionBusiness.java` and `SessionController.java`.

vip-portal/src/main/webapp/js/home.js
ethaaalpha: Okay, I've changed to `response.ok`. 
About the JS notes, I'll write then down into a ticket of @hippolyteblot to make sure these issues will be handled by the new front.

vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/controller/SessionController.java
```diff
@@ -1,22 +1,92 @@
...
+        try {
+            Session session = sessionBusiness.signin(credentials);
+            CsrfToken token = new CookieCsrfTokenRepository().generateToken(request);
+
+            response.addCookie(createCookie(CoreConstants.COOKIES_SESSION, session.id,
+                    CoreConstants.COOKIES_DURATION, true));
+            response.addCookie(createCookie(CoreConstants.COOKIES_USER, URLEncoder.encode(session.email, "UTF-8"),
+                    CoreConstants.COOKIES_DURATION, true));
+            response.addCookie(createCookie(
+                    CoreConstants.COOKIES_CSRF_TOKEN, token.getToken(), CoreConstants.COOKIES_DURATION, false));
```
axlbonnet: i still to review why this one is not httponly

axlbonnet (commented):
Seems good except for a few remarks.

Also, it would be nice to check if spring security login/session/cookie could replace some stuff here.

vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/controller/SessionController.java
```diff
@@ -1,22 +1,92 @@
...
-    @GetMapping("/session")
+
+    @GetMapping
     public Session getSession() {
-        Session session = new Session();
-        session.username = "test";
-        session.password = null;
-        session.level = UserLevel.Beginner;
-        return session;
+        return sessionBusiness.getSession();
```
axlbonnet: Maybe renew the cookies here to avoid them expiring after 1 week.

vip-portal/src/main/webapp/js/sign-in.js
```diff
@@ -99,40 +99,12 @@ function resetPassword() {
...
-function checkIfCookieExist(){
-    if (getCookie("vip-cookie-user") == true && getCookie("vip-cookie-user") == true) {
-        window.location.href="home.html";
-        return true;
-    } else {
-        return false;
-    }
+function checkSession() {
+    fetch("/internal/session").then(function (response) {
+        if ( ! response.ok) {
```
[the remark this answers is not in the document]
ethaaalpha: yep, it will work better :)

vip-api/src/main/java/fr/insalyon/creatis/vip/api/controller/OidcLoginController.java
```diff
@@ -65,9 +67,15 @@ public RedirectView getOauth2LoginInfo(
 
         try {
             User vipUser = configurationBusiness.getOrCreateUser((String) userAttributes.get("email"), domainName, null);
+            Session session = new Session();
+
+            session.id = vipUser.getSession();
+            session.email = vipUser.getEmail();
+            
             SecurityContextHolder.clearContext(); // destroy spring session and use VIP own session mechanism
-            vipSessionBusiness.setVIPSession(request, response, vipUser); // creates VIP cookies and session
```
axlbonnet: We just lost the call to `configurationBusiness.updateUserLastLogin(user.getEmail());` in the process.
Not sure where to put it although.

PR #600, opened 2025-10-07
implement /internal/applications endpoint

ethaaalpha:
This pull request depends on [/internal/session](https://github.com/virtual-imaging-platform/VIP-portal/pull/595) pull request.

What's new !

### Business Logic Changes
Implementation of the new "base" class of business components: `CommonBusiness`.
This class is supposed to be inherited **from every business** (the change will be progressive when will implement other endpoints). However, this class contain `ConfigurationBusiness`, for the moment that's the only "exception". (=_ConfigurationBusiness_ cannot inherit _CommonBusiness_ due to **circular problem**)

#### Permissions
The new `CorePermissions` (present in CommonBusiness) class implement few methods to check user roles and matching of privates users groups.

#### Annotations
Implementation of the `VipExternalSafe` annotation to mark methods that are considered "safe" to be used inside controller (=they perform permissions checks on returned values).  

### Application Controller
This new controller implement all of this methods:
- GET /applications
- GET/POST/UPDATE/DELETE /applications/{id}

The `GET /applications` support basic pagination and filtering using `group`, `offset` and `limit` query parameters. The implementation of pagination is made by the new `PageBuilder` and `PrecisePage` classes.

Related tests are also presents in `ApplicationControllerIT`.

#### Application permissions
| Create / Edit / Delete    | Yes | **only** applications in user **private groups** | No |
| Associate Group  | Yes | **only** at application **creation** | No |

### DIFF
--- a/vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/business/ApplicationBusiness.java
+++ b/vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/business/ApplicationBusiness.java
@@ -31,6 +31,19 @@
  */
 package fr.insalyon.creatis.vip.application.server.business;
 
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Transactional;
+
 import fr.insalyon.creatis.vip.application.client.bean.AppVersion;
 import fr.insalyon.creatis.vip.application.client.bean.Application;
 import fr.insalyon.creatis.vip.application.server.dao.ApplicationDAO;
@@ -45,15 +58,6 @@ import fr.insalyon.creatis.vip.core.server.business.base.CommonBusiness;
 import fr.insalyon.creatis.vip.core.server.dao.DAOException;
 import fr.insalyon.creatis.vip.core.server.inter.annotations.VIPExternalSafe;
 import fr.insalyon.creatis.vip.core.server.model.PrecisePage;
-
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.stereotype.Service;
-import org.springframework.transaction.annotation.Transactional;
-
-import java.util.*;
-import java.util.stream.Collectors;
 
 @Service
 @Transactional
@@ -78,9 +82,12 @@ public class ApplicationBusiness extends CommonBusiness {
     @VIPExternalSafe
     public void add(Application app) throws BusinessException {
         permissions.checkLevel(UserLevel.Administrator, UserLevel.Developer);
-        // this check is supposed to happend rarely since 
-        // we recommand associating groups after application creation (using update)
-        permissions.checkOnlyUserPrivateGroups(app.getGroups());
+
+        if (userSupplier.get().getLevel().equals(UserLevel.Developer)) {
+            // developers can only assign from private groups they belong to
+            // at application creation
+            permissions.checkOnlyUserPrivateGroups(app.getGroups());
+        }
         try {
             applicationDAO.add(app);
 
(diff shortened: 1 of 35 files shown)


vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/business/ApplicationBusiness.java
```diff
@@ -31,83 +31,162 @@
...
         this.appVersionBusiness = appVersionBusiness;
+        this.pageBuilder = pageBuilder;
     }
 
-    public void add(Application application) throws BusinessException {
+    @VIPExternalSafe
+    public void add(Application app) throws BusinessException {
+        permissions.checkLevel(UserLevel.Administrator, UserLevel.Developer);
+
+        if (userSupplier.get().getLevel().equals(UserLevel.Developer)) {
```
axlbonnet: You can add a `getUser()` method in CommonBusiness for simplicity.
Maybe even a `getUserLevel()` if it is needed at several places.

vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/controller/ApplicationController.java
```diff
@@ -1,17 +1,95 @@
...
+            throw new ApiException(ApiException.ApiError.INPUT_FIELD_NOT_VALID, id);
+        } else {
+            try {
+                Application existingApp = applicationBusiness.getApplication(id);
+                if (existingApp == null) {
+                    applicationBusiness.add(app);
+                } else {
+                    applicationBusiness.update(app);
+                }
+                return applicationBusiness.getApplication(id);
```
axlbonnet: what do you think of making `applicationBusiness.add` and `applicationBusiness.update` return the Application to avoid having to do an additional `applicationBusiness.getApplication` ?

vip-core/src/test/java/fr/insalyon/creatis/vip/core/integrationtest/database/BaseSpringIT.java
```diff
@@ -95,6 +103,13 @@ protected void assertRowsNbInTable(String tableName, int expectedNb) {
         assertEquals(expectedNb, rowsNb);
     }
 
+    protected User createUser(String email, UserLevel level) throws GRIDAClientException, BusinessException {
+        User u = createUser(email, UUID.randomUUID().toString().substring(0, 4));
+
+        configurationBusiness.updateUser(u.getEmail(), level, u.getCountryCode(), u.getMaxRunningSimulations(), false);
+        return configurationBusiness.getUserWithGroups(email);
```
axlbonnet: not sure to understand.
i think the user has no group here, so i do not get why there is `getUserWithGroups` ?

vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/controller/ApplicationController.java
```diff
@@ -1,17 +1,95 @@
...
+            }
+        } catch (BusinessException e) {
+            throw new ApiException(e);
+        }
+    }
+
+    @DeleteMapping(value = "{id}")
+    public void deleteApplication(@PathVariable String id) throws ApiException {
+        try {
+            applicationBusiness.remove(id);
```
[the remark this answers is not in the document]
ethaaalpha: mettre erreur

vip-application/src/test/java/fr/insalyon/creatis/vip/application/integrationtest/ApplicationControllerIT.java
```diff
@@ -0,0 +1,293 @@
...
+import fr.insalyon.creatis.vip.application.client.bean.Application;
+import fr.insalyon.creatis.vip.core.client.bean.Group;
+import fr.insalyon.creatis.vip.core.client.bean.GroupType;
+import fr.insalyon.creatis.vip.core.client.bean.User;
+import fr.insalyon.creatis.vip.core.client.view.user.UserLevel;
+import fr.insalyon.creatis.vip.core.integrationtest.database.BaseSpringIT;
+import fr.insalyon.creatis.vip.core.server.SpringInternalApiConfig;
+
+@ContextConfiguration(classes = { SpringInternalApiConfig.class })
+public class ApplicationControllerIT extends BaseSpringIT {
```
axlbonnet: You can extend `BaseInternalApiSpringIT`

PR #602, opened 2025-10-20
refactoring of exceptions classes

ethaaalpha:
This pull request depends on [/internal/application](https://github.com/virtual-imaging-platform/VIP-portal/pull/600) pull request.

What's new!?

### Exception refactoring
For simplifying the use of exceptions and avoid some useless `try/catch/rethrow` patterns, we have decided to simplify them.

First of all, we have deleted `BusinessException` and replaced by `VipException` it-self. Others child implementation like *DAOException*, *CoreException* might be deleted in the future and can be replaced by something new!

Instead of creating *child-classes* for specific cases with specifics errors codes you can implement your own `VipError` **enum**!
Indeed `VipException` can take an object that implement `VipError`.

`VipError`  is a new interface that mix up the old `VipError` and `VipErrorMessage` classes!
You can now easily implement your own specific errors with an error code associated (see `ApiError`) in the concerned module.

Since a lot of errors are common, the `DefaultError` *enum* provided some of them.

#### Scheme
A scheme might be the easiest way to understand the new structure!

```mermaid
classDiagram
direction LR
  class VipException {
    + VipError error
  }
  class VipError {
	  + getCode()
	  + getMessage()
  }
  class DefaultError {
	  NOT_FOUND(1000, "item not found")
  }
  class ApiError {
	  APP_NOT_FOUND(8006, "app not found")
  }
	<<Interface>> VipError
	<<Enumeration>> DefaultError
	<<Enumeration>> ApiError
  VipException --* VipError
  VipError --|> DefaultError
  VipError --|> ApiError
	class VipException:::Aqua
	class VipException:::Sky
	class VipError:::Peach
	class DefaultError:::Ash
	class ApiError:::Ash
	classDef Peach :,stroke-width:1px, stroke-dasharray:none, stroke:#FBB35A, fill:#FFEFDB, color:#8F632D
	classDef Aqua :,stroke-width:1px, stroke-dasharray:none, stroke:#46EDC8, fill:#DEFFF8, color:#378E7A
	classDef Sky :,stroke-width:1px, stroke-dasharray:none, stroke:#374D7C, fill:#E2EBFF, color:#374D7C
	classDef Ash :,stroke-width:1px, stroke-dasharray:none, stroke:#999999, fill:#EEEEEE, color:#000000
```

> [!NOTE]
> You can still use `VipException` without `VipError` by using others classical contructors

### Others changes
- `ApiException` (which became `ApiError`) is moved back into **vip-api** *module*.
[...]

PR #610, opened 2025-11-13
preparation of gwt migration

ethaaalpha:
This pull request depends on [refactoring of exceptions](https://github.com/virtual-imaging-platform/VIP-portal/pull/602) pull request.

The modifications in this PR are essentially preparing the future development according to the **GWT migration**.

#### New folder for models
All the models files were moved from `client/bean/**` to a new dedicated folder `models` in each module. 
This allows us to exclude GWT more easily from compilation for testing.

#### New maven profile
You can now exclude all client files from being compiled by using `mvn clean compile -P no-gwt`.  
In this case all the `**/client/**` and `**/server/rpc/**` are excluded from the compilation.
Also GWT won't run permutations.

This is will help us clean the code and prevent business dependencies from GWT code.


ethaaalpha (commented):
mainly imports changes, very few real codes changes. Merge using squash would be better

ethaaalpha:
@axlbonnet About the codeQL issue, it's GWT code related. Since it will be deprecated in few months do we try to fix the issue or not? 
EDIT: I've searched and the path is actually filtered by `isLFCPathAllowed` in `downloadByPath`. I think we can dismiss these alerts if you agree?

PR #640, opened 2026-02-23
implement /internal/appversions endpoint

ethaaalpha:
This pull request depend on [internal tags](https://github.com/virtual-imaging-platform/VIP-portal/pull/614) pull request.

What's new !

### Appversion endpoint
This new controller implement all "CRUD" methods (like **/internal/application** endpoint).  
About the permission this is almost the same deal than with `Application` objects.

#### Permissions
| Create / Edit / Delete    | Yes | **only** applications in user **private groups** | No |
| Associate Resources  | Yes | **only** at application **creation** | No |

> [!NOTE]
> `tags` related objects are manageable by admin and developers.


.vscode/settings.json
```diff
@@ -0,0 +1,6 @@
+{
+    // "editor.codeActionsOnSave": {
+    //     "source.fixAll": "explicit",
+    //     "source.organizeImports": "always",
+    // },
+}
```
axlbonnet: do not commit that

vip-application/src/main/java/fr/insalyon/creatis/vip/application/models/AppVersion.java
```diff
@@ -21,7 +22,8 @@ public class AppVersion implements IsSerializable {
     private String source;
     private String note;
 
-    public AppVersion() {}
+    public AppVersion() {
+    }
```
axlbonnet, line 27: Do we not associate objects (Tag, Ressources) with Set instead of List ?

vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/business/base/PermissionChain.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    private PermissionChain chain(CheckedRunnable<VipException> action, UserLevel required) {
+        items.add(new Scope(required, action));
+        return this;
+    }
+
+    // perform chain filtering (like ResponsabilityChain pattern)
+    // if user do not enter in any filter, it will be rejected
+    public void fitler() throws VipException {
```
axlbonnet: `fitler` typo

PR #641, opened 2026-02-24
fix jndi file loading

ethaaalpha:


axlbonnet:
Excellent ! Thanks for the improvement.

PR #642, opened 2026-02-24
implement /internal/groups endpoint

ethaaalpha:
This pull request depend [internal appversions](https://github.com/virtual-imaging-platform/VIP-portal/pull/640) pull request.

What's new!

### Groups endpoint

This new controller is very simple and implement all "CRUD" methods.

#### Permissions
Here is the permissions grid of the endpoint:

| Create / Edit / Delete    | Yes | No | No |


.vscode/settings.json
```diff
@@ -0,0 +1,6 @@
+{
+    // "editor.codeActionsOnSave": {
+    //     "source.fixAll": "explicit",
+    //     "source.organizeImports": "always",
+    // },
+}
```
axlbonnet: file to remove

vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/rpc/ConfigurationServiceImpl.java
```diff
@@ -159,7 +159,7 @@ public void removeGroup(String groupName) throws CoreException {
         try {
             authenticateSystemAdministrator(logger);
             trace(logger, "Removing group '" + groupName + "'.");
-            groupBusiness.remove(getSessionUser().getEmail(), groupName);
+            groupBusiness.remove(groupName);
```
axlbonnet, line 162: don't we have to set the SecurityContext with `putUserInSpringSecurityContext` as in ApplicationServiceImpl ?
(also in add / update / get)

ethaaalpha: as we discussed we use a generic method that define the context for each gwt request

PR #653, opened 2026-03-19
implement `/internal/users` endpoint

ethaaalpha:
What's new !

### Users endpoint
This new controller implement all "CRUD" methods (like **/internal/application** endpoint).  

> [!NOTE]
> `activation/password/reset-password` need to be implemented as sub-endpoints.

#### `JsonView` filtering
This pull request include a major change on how the JSON data is handled by controllers.
The `@JsonView` annotation allows us to easily exclude/filter some fields from being treated by the backend or sended back to the user.

To handle both `response` and `request` JSON filtering, I have implemented the two classes: `DataViewsResponseHandler` and `DataViewsRequestHandler`.

> [!WARNING]
> Make some tests when using these annotations, it can easily leads to data leaking!

#### Permissions
| Permission      | Admin | Dev / User |
|---------------|-------|------|
| Read                | Yes | only himself |
| Create / Edit / Delete    | Yes | only edit/delete himself | 
| Associate Groups  | Yes | only himself and public groups (can also leave private groups but no join) | 

#### Database modification
A new column have been added to `VIPUsers` table. It's `id`.  
You should run the migration script for migration. (available [here](https://github.com/virtual-imaging-platform/vip-scripts/tree/migration_uuid))

#### `ConfigurationBusiness` disappear!
To improve code clarity and cleanness I've splitted `ConfigurationBusiness` into multiples files: 
- _UserBusiness_
- _AuthenticationBusiness_
- _TermsOfUseBusiness_
- _ProxyBusiness_
- _PasswordBusiness_

#### new email class template
I've also created a new util class named `EmailTemplateUtils` to group and normalize email templates used inside the whole project.


vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/model/Session.java
```diff
@@ -1,7 +1,11 @@
 package fr.insalyon.creatis.vip.core.server.model;
 
+import com.fasterxml.jackson.annotation.JsonView;
+
 import fr.insalyon.creatis.vip.core.client.view.user.UserLevel;
+import fr.insalyon.creatis.vip.core.server.inter.DataViews;
 
+@JsonView(DataViews.User.class)
```
axlbonnet, line 8: Is it needed ?
I read in the javadoc that setting `@JsonView` on the class defines it for properties of this type, which is not useful in our case I thing.
Is it necessary to make SessionController work ?

vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/model/ErrorCodeAndMessage.java
```diff
@@ -1,7 +1,11 @@
 package fr.insalyon.creatis.vip.core.server.model;
 
+import com.fasterxml.jackson.annotation.JsonView;
+
+import fr.insalyon.creatis.vip.core.server.inter.DataViews;
 import jakarta.validation.constraints.NotNull;
 
+@JsonView(DataViews.User.class)
```
axlbonnet, line 8: Same as in Session.

vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/dao/mysql/UserData.java
```diff
@@ -152,46 +132,20 @@ public boolean activate(String email, String code) throws DAOException {
...
-                    + "last_login, level, country_code, max_simulations,termsUse,lastUpdatePublications,failed_authentications,account_locked "
-                    + "FROM VIPUsers "
-                    + "WHERE email=?");
-
+    public User get(String email) throws DAOException {
+        String query =  "SELECT "
+        +               "id, email, next_email, first_name, last_name, institution, "
+        +               "code, confirmed, folder, session, registration, "
+        +               "last_login, level, country_code, max_simulations, "
+        +               "termsUse, lastUpdatePublications, failed_authentications, account_locked, apikey "
```
axlbonnet: As the list of properties is the same is all methods, I suggest putting it in a constant in this class.

vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/dao/mysql/CoreDataInitializer.java
```diff
@@ -61,7 +63,8 @@ public void onStartup() {
 
     private void initializeUserTables() {
         if (tableInitializer.createTable("VIPUsers",
-                "email VARCHAR(255), "
+                          "id VARCHAR(" + CoreConstants.UUID_SIZE + "), "
```
axlbonnet, line 66: Also note the update in database_update.sql at the root of the project.

vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/business/ProxyBusiness.java
```diff
@@ -0,0 +1,43 @@
...
+    private final EmailBusiness emailBusiness;
+
+    @Autowired
+    public ProxyBusiness(ProxyClient proxyClient, Server server, EmailBusiness emailBusiness) {
+        this.server = server;
+        this.proxyClient = proxyClient;
+        this.emailBusiness = emailBusiness;
+    }
+
+    @Scheduled(fixedDelay = 10, timeUnit = TimeUnit.HOURS)
```
axlbonnet, line 28: Don't you need `@EnableScheduling` on the spring config class ?

vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/business/UserBusiness.java
```diff
@@ -0,0 +1,393 @@
...
+                // only admin can edit others accounts
+                throw new VipException(DefaultError.ACCESS_DENIED);
+            }
+            Set<Group> diff = new HashSet<>(user.getGroups());
+            diff.retainAll(getUser().getGroups()); // only examine modifications
+
+            // here user try to join private group (forbidden)
+            // but it's okay if the user try to leave a private group
+            if ( ! diff.stream().allMatch(Group::isPublicGroup)) {
+                System.err.println("PAR ICI");
```
axlbonnet: Il est passé par ici... il repassera par là...

vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/business/UserBusiness.java
```diff
@@ -0,0 +1,393 @@
...
+            if ( ! user.getId().equals(getUser().getId())) {
+                // only admin can edit others accounts
+                throw new VipException(DefaultError.ACCESS_DENIED);
+            }
+            Set<Group> diff = new HashSet<>(user.getGroups());
+            diff.retainAll(getUser().getGroups()); // only examine modifications
+
+            // here user try to join private group (forbidden)
+            // but it's okay if the user try to leave a private group
+            if ( ! diff.stream().allMatch(Group::isPublicGroup)) {
```
axlbonnet: not sure of the algorithm : diff contains the groups the user wants to stay in, so excludes the new ones. No ?

vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/business/UserBusiness.java
```diff
@@ -0,0 +1,393 @@
...
+                System.err.println("PAR ICI");
+                throw new VipException(DefaultError.ACCESS_DENIED);
+            }
+        }
+
+        try {
+            userDAO.update(user);
```
axlbonnet, line 128: With the user jsonview, this will not contained hidden things like `comfirmed` and `locked`, so it will erase them ?
Also it seems possible to change its own level ?

vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/business/ProxyBusiness.java
ethaaalpha: effectivement cela marche mieux avec!

vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/business/UserBusiness.java
ethaaalpha: yes I think it was very inefficient

PR #656, opened 2026-03-25
API group type selection

hippolyteblot:
You can now choose if you want to get only the `APPLICATION` type groups or `RESOURCE` types groups.
Here is the new route: GET internal/groups?onlyApplications=<bool>&onlyResources=<bool>
Here is three different cases :
- Both boolean at false -> returns everything
- One of them -> returns only the selected type
- Both boolean at true -> throws a bad parameters error


vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/business/ApplicationBusiness.java
```diff
@@ -221,6 +222,9 @@ public String getCitation(String name) throws VipException {
     public void associate(Application app, String groupName) throws VipException {
         app = getApplication(app.getName());
         Group group = groupBusiness.get(groupName);
+        if (group == null) {
+            throw new VipException(DefaultError.BAD_INPUT_FIELD, groupName, "Unknown group");
```
axlbonnet: I suggest using the `NOT_FOUND` error.
Ideally the error message would also contain the type missing, and not just "Item {} not found", but we can note to improve that latear

vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/business/GroupBusiness.java
```diff
@@ -102,21 +103,39 @@ public void update(String name, Group group) throws VipException {
...
             if (getUserLevel().equals(UserLevel.Administrator)) {
-                return groupDAO.get();
+                groups = groupDAO.get();
             } else {
-                return usersGroupsDAO.getUserGroups(getUser().getEmail()).keySet().stream().toList();
+                groups = usersGroupsDAO.getUserGroups(getUser().getEmail()).keySet().stream().toList();
+            }
+            if (onlyApplications && onlyResources) {
+                // Both filters requested = error
+                throw new VipException(DefaultError.BAD_PARAMETERS, "onlyApplications and onlyResources cannot be both true!");
```
axlbonnet: Check at the beginning, before doing any request.

PR #658, opened 2026-04-10
VIP input maps handling

Albireo-B:


vip-api/src/main/java/fr/insalyon/creatis/vip/api/business/ExecutionBusiness.java
```diff
@@ -162,7 +162,9 @@ private Execution getExecutionFromSimulation(Simulation s, boolean summarize) th
             s.getID(), currentUserProvider.get().getFolder());
         logger.debug("Execution has " + inputs.size() + " inputs ");
         for (InOutData iod : inputs) {
-            e.getInputValues().put(iod.getProcessor(), iod.getPath());
+            for (Map<String, Object> inputMap : e.getInputValues()) {
```
axlbonnet: i think `e.getInputValues()` is empty at this point.
Anyway we have to solve the workflows-db issue before (see the moteurlite PR).

vip-api/src/main/java/fr/insalyon/creatis/vip/api/business/ExecutionBusiness.java
```diff
@@ -287,25 +289,29 @@ public void updateExecution(Execution execution) throws VipException {
...
-        Map<String, String> inputMap = new HashMap<>();
+        List<Map<String, String>> inputMaps = new ArrayList<>();
+        Object resultsLocation = execution.getResultsLocation();
+        for (Map<String, Object> inputValuesMap : execution.getInputValues()) {
+            Map<String, String> inputMap = new HashMap<>();
+            for (Entry<String, Object> restInput : inputValuesMap.entrySet()) {
+                inputMap.put(
+                        restInput.getKey(),
+                        handleRestParameter(restInput.getKey(), restInput.getValue()));
+            }
```
axlbonnet, line 331: if there are several inputs objets, then we must verify all inputs have only 1 element.

vip-api/src/main/java/fr/insalyon/creatis/vip/api/business/ExecutionBusiness.java
```diff
@@ -373,12 +379,15 @@ private String initExecution(String pipelineId,
...
-            if (inputValues.get(pp.getName()) != null) {
+            if (inputValues.stream().allMatch(inputMap -> inputMap.containsKey(pp.getName()))) {
                 continue;
             }
             // then ok if input has a default value (and we set it)
             if (pp.getDefaultValue() != null) {
-                inputValues.put(pp.getName(), pp.getDefaultValue().toString());
+                for (Map<String, String> inputMap : inputValues) {
+                    inputMap.put(pp.getName(), pp.getDefaultValue().toString());
+                }
```
axlbonnet: I think it does not handle the case if some input object have a value and some not.

vip-api/src/main/java/fr/insalyon/creatis/vip/api/model/Execution.java
```diff
@@ -13,7 +13,7 @@ public class Execution {
     private int timeout;
     private ExecutionStatus status;
     @NotNull
-    private Map<String, java.lang.Object> inputValues;
+    private List<Map<String, Object>> inputValues;
```
axlbonnet: I would like to not break the API with this change, and support an object as before, and a list.
I see 2 solutions, but there may be others:
- have inputValues as `Object` and have methods in Execution that help manage the different cases.
- implement specific Execution Jackson Serializer and Deserializer.

axlbonnet (commented):
Some things to discuss.

vip-api/src/main/java/fr/insalyon/creatis/vip/api/model/Execution.java
Albireo-B: Created InputValuesDeserializer which handles deserializing to List<Map<String,Object>>

vip-api/src/main/java/fr/insalyon/creatis/vip/api/business/ExecutionBusiness.java
Albireo-B: Indeed, modified to handle only missing key values

vip-api/src/main/java/fr/insalyon/creatis/vip/api/business/ExecutionBusiness.java
Albireo-B: Added an exception throw

vip-api/src/main/java/fr/insalyon/creatis/vip/api/business/ExecutionBusiness.java
Albireo-B: Some null checks were introduced by the latest PR

vip-api/src/main/java/fr/insalyon/creatis/vip/api/business/ExecutionBusiness.java
axlbonnet: I find it more logic to do the test before the loop, like:
```
if (pp.getDefaultValue() != null) {
  mapsWithoutKey.stream.forEach(...)
  continue;
}
 // then ok if it is optional
if (pp.isOptional()) {
  continue;
}
// error : pp is an empty input with no default value and it is not optional
logger.error("Error initialising {}, missing {} parameter", pipelineId, pp.getName());
throw new VipException(ApiError.INPUT_FIELD_MISSING, pp.getName());
```

Albireo-B: yes it makes sense

vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/business/simulation/parser/InputFileParser.java
```diff
@@ -100,33 +102,58 @@ public Map<String, String> handleXML(File file) throws VipException {
...
-            return inputs;
+            return Collections.singletonList(inputs);
         } catch (IOException | SAXException | ParserConfigurationException e) {
             logger.error("Error parsing {}", file.getName(), e);
             throw new VipException(e);
         }
     }
 
-    public Map<String, String> handleJSON(File file) throws VipException {
+    public List<Map<String, String>> handleJSON(File file) throws VipException {
```
axlbonnet, line 111: we'll make this one more simpler, and make it return the same thing as in the json, with tha `pathHandler` applied.
In other words:
- if it's a array keep it as a list
- if it's an object, create a list with only the object

iterate through all the objects item to apply `handleList`/`pathHandler` as before.

vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/business/WorkflowBusiness.java
```diff
@@ -452,7 +455,7 @@ public void purge(String simulationID) throws VipException {
         }
     }
 
-    public Map<String, String> relaunch(String simulationID, String currentUserFolder) throws VipException {
+    public List<Map<String, String>> relaunch(String simulationID, String currentUserFolder) throws VipException {
```
axlbonnet: We'll keep it simple:
- get the `List<Map<String, String>>` from `getInputFileParser`
- if there is only 1 element, return the `Map<String, String>`
- if there is more than 1, we do not support that yet, throw an exception

That way, this will still return `Map<String, String>` and won't change GWT stuff.

vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/business/util/ReproVipInputsParser.java
```diff
@@ -36,8 +36,8 @@ public ReproVipInputsParser(ExternalPlatformBusiness externalPlatformBusiness, S
         this.providerInformations.put("vip_url", vipURL);
     }
 
-    public void parse(Map<String, String> inputs) throws VipException {
-        simplifiedInputs = getExpandedInputs(inputs);
+    public void parse(List<Map<String, String>> inputsList) throws VipException {
+        simplifiedInputs = getExpandedInputs(inputsList.getFirst());
```
axlbonnet, line 44: Throw an exeption if there is more than 1 element.

axlbonnet (commented):
I think we can make relaunch a bit simpler, as we won't support relaunching execution with a list of inputs (this is API specific anyway, and the relaunch is UI-specific).

vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/business/simulation/parser/InputFileParser.java
Albireo-B: I made the changes, but this way :
- we lose the compatibility when trying to relaunch a new format execution (multiple input maps)
- the inputs property is not used in this function anymore

PR #664, opened 2026-04-16
adding /users/me

hippolyteblot:


vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/controller/UserController.java
```diff
@@ -42,6 +46,18 @@ public PrecisePage<User> list(@RequestParam(defaultValue = "0") @PositiveOrZero
...
+    public User getCurrentUser() throws VipException {
+        User currentUser = userProvider.get();
+        User user = userBusiness.get(currentUser.getId());
+
+        if (user == null) {
+            throw new VipException(DefaultError.NOT_FOUND, currentUser.getId());
+        } else {
+            return user;
+        }
+    }
```
axlbonnet, line 50: As all the logic about authorization and stuff is in the Business, I prefer to add the `getCurrentUser` method in UserBusiness.
For instance, if we want to change a behavior and empty a field from User, it will be more coherent as all user management is done in the same Business.

vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/controller/UserController.java
```diff
@@ -42,6 +46,18 @@ public PrecisePage<User> list(@RequestParam(defaultValue = "0") @PositiveOrZero
     return userBusiness.getAll(offset, quantity);
     }
 
+    @GetMapping(value = "me")
+    public User getCurrentUser() throws VipException {
+        User currentUser = userProvider.get();
+        User user = userBusiness.get(currentUser.getId());
```
axlbonnet: is it necessary to query again the same user ?

PR #669, opened 2026-04-22
fix endDate

MayssaRouissi:


vip-api/src/main/java/fr/insalyon/creatis/vip/api/business/ExecutionBusiness.java
```diff
@@ -152,19 +153,18 @@ private Execution getExecutionFromSimulation(Simulation s, boolean summarize) th
...
                 s.getStatus() == null ? null : convertVIPtoCarminStatus(s.getStatus()),
-                null, // study identifier (not available in VIP yet)
-                null,// error codes and mesasges (not available in VIP yet)
-                s.getDate().getTime(),
-                null,// last status modification date (not available in VIP yet)
-                null// results location (not available in VIP yet)
+                null, //study identifier (not available in VIP yet)
+                null, //  error codes and mesasges (not available in VIP yet)
+                s.getDate().getTime(), // startDate
+                s.getEndDate() != null ? ((Date) s.getEndDate()).getTime() : null, //endDate
```
axlbonnet, line 161: i dont think the cast to `(Date)` is necessary.
`s.getEndDate() != null ? s.getEndDate().getTime() : null` should work.

axlbonnet (commented):
Good work !
need to merge develop and deal with conflict though.

vip-api/src/main/java/fr/insalyon/creatis/vip/api/business/ExecutionBusiness.java
MayssaRouissi: when I don’t cast it, it generates an error

axlbonnet:
Replaced by #681

PR #670, opened 2026-04-23
Add user register

hippolyteblot:
# Password Support for User Signup

This PR fixes account creation by properly supporting password handling during signup

## What changed

- Added a dedicated signup DTO (`SignUpUser`) used only for user registration
- Updated `SignUpForm` (already present) to use this signup specific DTO
- Logs around the signup flow to make debugging and maintenance easier
- `@JsonIgnore` on `User.password` to prevent password exposure through the default `User` model


vip-core/src/main/java/fr/insalyon/creatis/vip/core/models/User.java
```diff
@@ -41,6 +41,7 @@ public class User implements IsSerializable {
     private String code;
     private String folder;
     private String session;
+    @JsonIgnore
```
axlbonnet: did you have a case when the password was visible ?

vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/controller/dto/SignUpForm.java
```diff
@@ -1,12 +1,12 @@
...
 import com.fasterxml.jackson.annotation.JsonView;
 
 import fr.insalyon.creatis.vip.core.server.inter.DataViews;
 
 @JsonView(DataViews.User.class)
 public class SignUpForm {
- 
-    public User user;
+
+    public SignUpUser user;
```
axlbonnet: I do not like extending User with a dummy class that just override the password getter and setter
I suggest:
- getting rid of SignUpUser
- keeping User here
- adding a password field in SignUpForm
- moving SignUpForm it the models package
- eventually renaming SignUpForm to `UserAndPassword`

vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/business/AuthenticationBusiness.java
```diff
@@ -146,6 +149,9 @@ public User signup(User user, String comments, boolean automaticCreation, boolea
                 groupsString.append(group.getName()).append(", ");
             }
 
+                logger.info("Signup persistence succeeded for email='{}' with generatedId='{}'",
```
axlbonnet: weird indentation

vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/business/AuthenticationBusiness.java
```diff
@@ -162,11 +168,13 @@ public User signup(User user, String comments, boolean automaticCreation, boolea
                         false, user.getEmail());
             }
 
+            logger.info("Signup flow completed for email='{}'", user.getEmail());
             return user;
         } catch (GRIDAClientException | UnsupportedEncodingException | NoSuchAlgorithmException ex) {
             logger.error("Error signing up user {}", user.getEmail(), ex);
             throw new VipException(ex);
         } catch (DAOException ex) {
+            logger.error("DAO error while signing up user {}", user.getEmail(), ex);
```
axlbonnet: normally the dao error is already logged when the DAOException is raised.

axlbonnet (commented):
I think it would be better to not extend User.

vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/controller/UserController.java
```diff
@@ -70,11 +75,33 @@ public User update(@PathVariable String id, @RequestBody @Valid User user) throw
...
-    public User create(@RequestBody @Valid SignUpForm form) throws VipException {
+    public User create(@RequestBody @Valid UserAndPassword form) throws VipException {
+        final String email = form != null && form.user != null ? form.user.getEmail() : null;
+        final String countryCode = form != null && form.user != null && form.user.getCountryCode() != null
+                ? form.user.getCountryCode().name()
+                : null;
+        final boolean hasPassword = form != null && form.password != null && !form.password.isBlank();
+        final int groupsCount = form != null && form.user != null && form.user.getGroups() != null
+                ? form.user.getGroups().size()
+                : -1;
```
axlbonnet: a bit ugly just for logs.
i think that an annotation (`@NotNull` ?) should be enough to ensure form is not null
also an annotation on `UserAndPassword::user` should ensure it's not null.

PR #675, opened 2026-05-06
Add user validation

hippolyteblot:
# User validation feature
1. Creating a PUT users/'id'/activate, with the validation code in the body that allow to validate an account.
2. Adding a new field 'confirmed' in the Session model, to give to the frontend a way to know if the user is already validate after login.


vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/controller/UserController.java
```diff
@@ -97,4 +98,9 @@ public User create(@RequestBody @Valid UserAndPassword form) throws VipException
         logger.info("Signup completed: email='{}', generatedId='{}'", createdUser.getEmail(), createdUser.getId());
         return createdUser;
     }
-}
\ No newline at end of file
+
+    @PutMapping(value = "{id}/activate")
+    public void activate(@PathVariable String id, @RequestBody @Valid ActivationForm form) throws VipException {
+        authenticationBusiness.activate(id, form.getCode());
```
axlbonnet: `authenticationBusiness.activate` takes an email as 1st param, not the id.

vip-core/src/main/java/fr/insalyon/creatis/vip/core/models/ActivationForm.java
```diff
@@ -0,0 +1,28 @@
+package fr.insalyon.creatis.vip.core.models;
+
+import com.fasterxml.jackson.annotation.JsonView;
+import com.google.gwt.user.client.rpc.IsSerializable;
+import fr.insalyon.creatis.vip.core.server.inter.DataViews;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+
+@JsonView(DataViews.User.class)
+public class ActivationForm implements IsSerializable {
```
axlbonnet: Not a fan of "xxxxForm" name.
I prefer `ActivationCode` in this case.
Also add a @NotBlanc annotation on the `code` field.

vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/controller/UserController.java
```diff
@@ -97,4 +98,10 @@ public User create(@RequestBody @Valid UserAndPassword form) throws VipException
         logger.info("Signup completed: email='{}', generatedId='{}'", createdUser.getEmail(), createdUser.getId());
         return createdUser;
     }
-}
\ No newline at end of file
+
+    @PutMapping(value = "{id}/activate")
+    public void activate(@PathVariable String id, @RequestBody @Valid ActivationCode activationCode) throws VipException {
+        User user = userBusiness.get(id);
```
axlbonnet, line 104: throw a little exception if user is null.

PR #676, opened 2026-05-18
Password update

hippolyteblot:
Handel the password update for the current user. Admin are not able to update it

### DIFF
--- a/vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/business/UserBusiness.java
+++ b/vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/business/UserBusiness.java
@@ -27,6 +27,7 @@ import fr.insalyon.creatis.vip.core.client.view.user.UserLevel;
 import fr.insalyon.creatis.vip.core.client.view.util.CountryCode;
 import fr.insalyon.creatis.vip.core.models.Group;
 import fr.insalyon.creatis.vip.core.models.User;
+import fr.insalyon.creatis.vip.core.models.UserAndPassword;
 import fr.insalyon.creatis.vip.core.server.business.base.CommonBusiness;
 import fr.insalyon.creatis.vip.core.server.dao.DAOException;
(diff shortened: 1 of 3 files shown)


vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/business/UserBusiness.java
```diff
@@ -329,6 +333,28 @@ public User getUserWithSession(String email) throws DAOException {
         return userDAO.get(email);
     }
 
+    @VIPExternalSafe
+    public void updateUserPassword(String userId, UserAndPassword form) throws VipException {
+        if (form == null || form.password == null || form.password.isBlank()) {
+            throw new VipException(DefaultError.BAD_INPUT_FIELD, "password", "Password is required!");
+        }
+
+        String targetEmail = userId;
```
axlbonnet: weird, i think that id should be the unique id, not the email.

vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/business/UserBusiness.java
```diff
@@ -329,6 +333,28 @@ public User getUserWithSession(String email) throws DAOException {
...
+            targetEmail = form.user.getEmail();
+        }
+
+        User currentUser = getCurrentUser();
+        if (!currentUser.getEmail().equals(targetEmail)) {
+            throw new VipException(DefaultError.ACCESS_DENIED);
+        }
+
+        passwordBusiness.setPassword(targetEmail, form.password);
+    }
```
axlbonnet, line 349: I prefer verifying the first stuff (that `userId` is the same as `form.user.id`) in the controller. Also that password is provided (but the annotation should ensure that).
Then this method will receive id + password, ensure the id is the current user and go on.
Maybe verify somewhere the password is not empty/blank (or change the validation annotation)

PR #677, opened 2026-05-21
Endpoint /internal/messages

hippolyteblot:
This new PR allows the backend to provide new routes under `/internal/messages`.

Here are the new endpoints:
- GET `/internal/messages`: Returns inbox messages for the current user
- GET `/internal/messages/send`: Returns messages sent by the current user
- GET `/internal/messages/groups`: Returns inbox group messages for the current user
- POST `/internal/messages`: Allows the current user to send a new message (can be a group message)
- DELETE `/internal/messages/{id}`: Deletes a message for the current recipient only
- DELETE `/internal/messages/send/{id}`: Deletes a message as the sender (removes the message for all recipients)
- PUT `/internal/messages/{id}/read`: Marks the message as read by the current user (must be a recipient)

Here is also two new endpoints (and associated business methods) that allows to search for a user, or a group (used for recommendations when sending messages)


vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/business/UserBusiness.java
```diff
@@ -181,6 +182,38 @@ public List<User> getUsers() throws VipException {
         }
     }
 
+    @VIPExternalSafe
+    public List<User> searchUsers(String query, int limit) throws VipException {
```
axlbonnet, line 186: to avoid making available all the user information to every body we'll do 2 things:
- as in GWT, making this method (and message writing) forbidden to the BEGINNER role
- "prunning" the user to keep only : id / first name / last name

vip-social/src/main/java/fr/insalyon/creatis/vip/social/server/controller/MessageController.java
```diff
@@ -0,0 +1,110 @@
...
+
+    @DeleteMapping("/{id}")
+    public void deleteMessage(@PathVariable long id) throws VipException {
+        User currentUser = userBusiness.getCurrentUser();
+        messageBusiness.removeByReceiver(id, currentUser.getEmail());
+    }
+
+    @DeleteMapping("/send/{id}")
+    public void deleteSentMessage(@PathVariable long id) throws VipException {
+        messageBusiness.remove(id);
```
axlbonnet, line 87: add in MessageBusiness the verification the message is sent by the current user (or the current user is an admin)

vip-social/src/main/java/fr/insalyon/creatis/vip/social/models/Message.java
axlbonnet, line 34: we'll also use this model for POST with the following fields read-only :
- id
- sender
- posted
- postedDate
- read

vip-social/src/main/java/fr/insalyon/creatis/vip/social/models/GroupMessage.java
axlbonnet, line 33: same as message, make the folowing fields read-only:
- id
- sender
- posted
- postedDate

vip-social/src/main/java/fr/insalyon/creatis/vip/social/server/controller/MessageController.java
```diff
@@ -0,0 +1,110 @@
...
+                    messageBusiness.sendGroupMessage(
+                            currentUser,
+                            groupName,
+                            userBusiness.getUsersFromGroup(groupName),
+                            request.getSubject(),
+                            request.getMessage());
+                }
+            }
+            return;
+        }
```
axlbonnet, line 75: We will make 2 different endpoints:
- POST /messages (with a Message model)
- POST /messages/groups (with a GroupMessage model)

get rid of SendMessageRequest

Simple verification:
- beginner cannot post on /messages
- only a group admin of the group specified on the group message (or a VIP admin) can post on /messages/groups

Maybe validation of NotBlank through @NotBlank annotation

vip-social/src/main/java/fr/insalyon/creatis/vip/social/server/business/MessageBusiness.java
```diff
@@ -76,31 +80,37 @@ public List<GroupMessage> getGroupMessages(String groupName, Date startDate)
...
-            messageDAO.removeByReceiver(id, receiver);
+            User currentUser = userBusiness.getCurrentUser();
+            messageDAO.removeByReceiver(id, currentUser.getEmail());
         } catch (DAOException ex) {
             throw new VipException(ex);
         }
     }
 
     public void removeGroupMessage(long id) throws VipException {
+        assertCurrentUserCanDeleteGroupMessage(id);
```
axlbonnet, line 112: we'll do simpler and keep the same right as `assertCurrentUserCanSendGroupMessage`.
all group admins can delete all messages.

vip-social/src/main/java/fr/insalyon/creatis/vip/social/server/controller/MessageController.java
```diff
@@ -0,0 +1,99 @@
...
+
+        if (message == null || message.getReceivers() == null || message.getReceivers().length == 0) {
+            throw new VipException(DefaultError.BAD_INPUT_FIELD, "receivers", "Receivers are required!");
+        }
+
+        // convert User[] receivers to String[] emails
+        String[] recipients = java.util.Arrays.stream(message.getReceivers())
+                .filter(r -> r != null && r.getEmail() != null && !r.getEmail().isBlank())
+                .map(r -> r.getEmail())
+                .toArray(String[]::new);
```
axlbonnet, line 62: we may have an issue here.
the searchUser endpoint do not return email, but ids.
i think receivers here should be identified by their ids, not their emails.

vip-social/src/main/java/fr/insalyon/creatis/vip/social/server/business/MessageBusiness.java
axlbonnet: i do not see that

vip-social/src/main/java/fr/insalyon/creatis/vip/social/server/business/MessageBusiness.java
```diff
@@ -122,6 +136,16 @@ public void sendMessage(
...
+            } else {
+                // Convert recipient IDs to emails
+                List<String> emails = new ArrayList<>();
+                for (String recipientId : recipients) {
+                    User recipient = userBusiness.get(recipientId);
+                    if (recipient != null && !recipient.isAccountLocked()) {
+                        emails.add(recipient.getEmail());
+                    }
+                }
+                recipients = emails.toArray(new String[]{});
```
axlbonnet, line 162: This is still used from GWT where email are used I think.
I suggest testing if the first element is an email (with an `@`)

vip-social/src/main/java/fr/insalyon/creatis/vip/social/server/business/MessageBusiness.java
axlbonnet: Actually i'm fine with the current stuff

PR #678, opened 2026-05-26
update database

MayssaRouissi:
To make sure that the columns are initialized to ' '


axlbonnet:
I think we should ensure `note` can be null instead of making them blank by default.

database_update.sql
```diff
@@ -46,8 +46,9 @@ USE workflowsdb;
 ALTER TABLE Workflows ADD tags VARCHAR(255);
 
 -- v4.7
-ALTER TABLE VIPApplications ADD note TEXT;
-ALTER TABLE VIPAppVersions ADD note TEXT;
+ALTER TABLE VIPApplications ALTER COLUMN note DROP NOT NULL,ALTER COLUMN note SET DEFAULT NULL;
+ALTER TABLE VIPAppVersions ALTER COLUMN note DROP NOT NULL,ALTER COLUMN note SET DEFAULT NULL;
```
axlbonnet: This file is meant to contains the SQL requests for migration.
So for 4.7, the requests to create the note fields.
I get that the original requests were "fine", and now that the code support "null" values, we can keep the original requests.

PR #681, opened 2026-06-04
end date

MayssaRouissi:


vip-application/src/main/java/fr/insalyon/creatis/vip/application/models/Simulation.java
```diff
@@ -97,5 +99,12 @@ public String toString() {
     public void setSimulationName(String simulationName) {
         this.simulationName = simulationName;
     }
+      public Date getEndDate() {
```
axlbonnet: there seems to be a small indentation issue here.

vip-api/src/main/java/fr/insalyon/creatis/vip/api/business/ExecutionBusiness.java
```diff
@@ -155,11 +156,11 @@ private Execution getExecutionFromSimulation(Simulation s, boolean summarize) th
...
                 s.getStatus() == null ? null : convertVIPtoCarminStatus(s.getStatus()),
-                null, // study identifier (not available in VIP yet)
-                null,// error codes and mesasges (not available in VIP yet)
-                s.getDate().getTime(),
-                null,// last status modification date (not available in VIP yet)
-                null// results location (not available in VIP yet)
+                null, //study identifier (not available in VIP yet)
+                null, //  error codes and mesasges (not available in VIP yet)
+                s.getDate().getTime(), // startDate
+                s.getEndDate() != null ? ((Date) s.getEndDate()).getTime() : null,
```
axlbonnet: You still have an error with `s.getEndDate() != null ? s.getEndDate().getTime() : null` ?
Because normally `s.getEndDate()` is of type `Date` and it should be fine.
Can you show me you error ?

MayssaRouissi: just tested it again, and it worked without the cast. So the error I'm talking about must be a different one; I must have confused the two errors

PR #683, opened 2026-06-19
Fix back for front

hippolyteblot:
# Internal REST API fixes

1. Fixed the security filter in `InternalSecurityConfig.java` to allow non-authenticated users to activate their accounts.
2. Updated the "Get Groups" endpoint to return all groups available to the user, not only the groups they belong to.

### DIFF
--- a/vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/business/GroupBusiness.java
+++ b/vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/business/GroupBusiness.java
@@ -4,6 +4,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -114,7 +115,10 @@ public class GroupBusiness extends CommonBusiness {
             if (getUserLevel().equals(UserLevel.Administrator)) {
                 groups = groupDAO.get();
             } else {
-                groups = usersGroupsDAO.getUserGroups(getUser().getEmail()).keySet().stream().toList();
+                groups = Stream.concat(
+                    usersGroupsDAO.getUserGroups(getUser().getEmail()).keySet().stream(),
+                    groupDAO.get().stream().filter(Group::isPublicGroup)
+                ).distinct().toList();
             }
             if (onlyApplications) {
                 groups = groups.stream().filter((g) -> g.getType().equals(GroupType.APPLICATION)).toList();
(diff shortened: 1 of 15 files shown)


vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/business/GroupBusiness.java
```diff
@@ -114,7 +115,10 @@ public List<Group> get(boolean onlyApplications, boolean onlyResources) throws V
             if (getUserLevel().equals(UserLevel.Administrator)) {
                 groups = groupDAO.get();
             } else {
-                groups = usersGroupsDAO.getUserGroups(getUser().getEmail()).keySet().stream().toList();
+                groups = Stream.concat(
+                    usersGroupsDAO.getUserGroups(getUser().getEmail()).keySet().stream(),
+                    groupDAO.get().stream().filter(Group::isPublicGroup)
+                ).distinct().toList();
```
axlbonnet, line 121: You can use `GroupBusiness::getOrLoadUserGroups()` to avoid fetching them again.

vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/controller/UserController.java
```diff
@@ -117,11 +124,22 @@ public User create(@RequestBody @Valid UserAndPassword form) throws VipException
...
     @PutMapping(value = "{id}/activate")
-    public void activate(@PathVariable String id, @RequestBody @Valid ActivationCode activationCode) throws VipException {
-        User user = userBusiness.get(id);
-        if (user == null) {
-            throw new VipException(DefaultError.NOT_FOUND, User.class.getSimpleName(), id);
+    public Session activate(@PathVariable String id, @RequestBody @Valid ActivationCode activationCode,
+            HttpServletRequest request, HttpServletResponse response) throws VipException {
+        // the endpoint is public (permitAll), so we can't use
+        // userBusiness.get() that requires an authenticated user.
+        authenticationBusiness.activate(id, activationCode.getCode());
```
axlbonnet: Think you can make `authenticationBusiness.activate` return getUserWithSession, and also remove the fact that it creates the trash folder.

vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/controller/UserController.java
```diff
@@ -117,11 +124,22 @@ public User create(@RequestBody @Valid UserAndPassword form) throws VipException
...
+
+        // Auto-login after successful activation
+        try {
+            User user = userBusiness.getUserWithSession(id);
+            Session session = sessionBusiness.getSession(user);
+            sessionBusiness.createLoginCookies(request, response, session);
+            userBusiness.updateUserLastLogin(user.getEmail());
+            return session;
+        } catch (Exception e) {
```
axlbonnet: no umbrella catch

PR #688, opened 2026-07-02
Frontend prototype

hippolyteblot:


frontend/src/views/LandingView.vue
```diff
@@ -35,57 +38,33 @@ async function loadPublicApplications() {
...
+async function loadPublicPublications() {
+  try {
+    topPublications.value = await publicationsApi.getPublic()
+  } catch (error) {
+    console.error('Failed to load public publications:', error)
+    topPublications.value = []
+  }
+}
+
+import { getFrontendBase } from '@/utils/path'
```
axlbonnet: do you prefer putting the import here and not at the top ?

frontend/.env.development
```diff
@@ -0,0 +1,2 @@
+# Backend URL for the Vite dev proxy
+#VITE_BACKEND_URL=
```
axlbonnet: so each developer has to put its local VM url in this file ?
Should it NOT be in git ?

frontend/src/components/launch/FileInputControl.vue
```diff
@@ -51,15 +51,21 @@ async function onUploadChange(event: Event) {
...
 
   isUploading.value = true
   try {
     await ensureDirectory(userHome, 'direct-uploads')
-    await filesApi.createDirectory(`${userHome}/direct-uploads`, `${timestamp}`)
-    await filesApi.uploadFile(`${userHome}/direct-uploads/${timestamp}`, file)
-    const uploadedPath = `${userHome}/direct-uploads/${timestamp}/${file.name}`
+    await filesApi.createDirectory(`${userHome}direct-uploads`, `${timestamp}`)
+    await filesApi.uploadFile(`${userHome}direct-uploads/${timestamp}`, file)
+    const uploadedPath = `${userHome}direct-uploads/${timestamp}/${file.name}`
```
axlbonnet, line 68: Is there mechanism to clean up this directory ?
If not, we should think about how to do that (in a future PR)

frontend/src/views/ProfileView.vue
```diff
@@ -271,27 +273,50 @@ async function submitPasswordChange() {
...
     .catch(() => notificationsStore.error('Unable to copy API key.'))
 }
 
-function regenerateApiKey() {
+async function fetchApiKey() {
+  isLoadingApiKey.value = true
+  try {
+    apiKey.value = await apikeyApi.get()
+  } catch {
+    apiKey.value = null
```
axlbonnet, line 291: why not a notification error ?

frontend/src/views/ProfileView.vue
```diff
@@ -271,27 +273,50 @@ async function submitPasswordChange() {
...
+  } finally {
+    isLoadingApiKey.value = false
```
axlbonnet: Are the boolean `isLoadingApiKey` / `isRegeneratingApiKey` / `isDeletingApiKey` used somewhere ?

frontend/src/components/launch/FileInputControl.vue
hippolyteblot: No, there isn't. I was wondering that we are not going to remove uploaded file for reproducibility/tracability purposes, but maybe it's a bad behaviour for the storage size.

frontend/.env.development
hippolyteblot: Yes that is. I added it to show how to fill it, but I can remove it

hippolyteblot: Template file

PR #690, opened 2026-07-16
Feature/add engine route UI

bergamote156:


frontend/src/components/layout/Sidebar.vue
```diff
@@ -55,6 +56,11 @@ const mainNav = computed<NavItem[]>(() => [
     to: '/messages',
     icon: Mail,
   },
+  {
```
hippolyteblot, line 63: As Axel said, a small demarcation for admin functions in the sidebar would be good

vip-application/src/main/java/fr/insalyon/creatis/vip/application/client/rpc/ApplicationService.java
```diff
@@ -69,6 +69,8 @@ public static ApplicationServiceAsync getInstance() {
     public void removeEngine(String engineName) throws VipException;
     
     public List<Engine> getEngines() throws VipException;
+
+    public Engine getEngine(String engineName) throws VipException;
```
axlbonnet: As this is only used by GWT and it's used on the front-side of GWT it is not necessary.
I suggest removing that (and `ApplicationServiceImpl` and `ApplicationServiceAsync`)

vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/business/EngineBusiness.java
```diff
@@ -76,12 +98,12 @@ public List<Engine> getUsableEngines(Resource resource) throws VipException {
         engines = getByResource(resource);
```
axlbonnet: i think you can remove the redundant call above.

vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/business/EngineBusiness.java
```diff
@@ -76,12 +98,12 @@ public List<Engine> getUsableEngines(Resource resource) throws VipException {
...
 
         if (engines.isEmpty()) {
-            engines = get();
+            engines = getAll();
```
axlbonnet: seems not right, i think we should return an empty list here.

vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/business/EngineBusiness.java
```diff
@@ -91,7 +113,7 @@ public Engine selectEngine(List<Engine> availableEngines) throws VipException {
         try {
             for (Engine engine : availableEngines) {
                 long runningWorkflows = workflowDAO.getNumberOfRunningPerEngine(engine.getEndpoint());
-                if (runningWorkflows < min && ! engine.getEndpoint().isEmpty()) {
+                if (runningWorkflows < min && !engine.getEndpoint().isEmpty()) {
```
axlbonnet, line 111: i like putting a space after a negation for clarity, but i can be convinced to remove it :)

vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/business/EngineBusiness.java
```diff
@@ -47,16 +55,30 @@ public void update(Engine engine) throws VipException {
...
+
+        try {
+            return engineDAO.getAll();
+        } catch (DAOException ex) {
+            throw new VipException(ex);
+        }
+    }
+
+    public Engine get(String name) throws VipException {
         try {
```
axlbonnet, line 80: `get` and `getAll` are used in the controller and should have `@VIPExternalSafe` + `permissions.filter(PermissionChain::admin);`

But as `getAll` is used in `getUsableEngines` which is an internal call that must not be `@VIPExternalSafe` we have an issue.
I suggest calling directly `engineDAO.getAll` in `getUsableEngines`, or creating another `getAllInternal` for internal usage (but I find that ugly).
The current architecture with business's mixing external and internal stuff is akward. I think we should separate each business in 2. Have you an opinion ?

frontend/src/components/layout/Sidebar.vue
bergamote156: Fixed using user role matching filter and separator + distinct navbar

vip-application/src/main/java/fr/insalyon/creatis/vip/application/client/rpc/ApplicationService.java
bergamote156: Removed getEngine from rpc interfaces.

vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/business/EngineBusiness.java
bergamote156: Removed completely as the DAO returns an empty list anyway if none is found.

vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/business/EngineBusiness.java
bergamote156: Was not intended :)
I think we should add a global auto-format to prevent such diff in the future.

vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/business/EngineBusiness.java
bergamote156: getUsableEngines already calls a distinct dao (getByResource), thus get and getAll are kept "externally safe". I did not add the annotation as it is filtered at the controller level in the security config.

vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/security/common/SpringPrincipalUser.java
```diff
@@ -26,7 +27,7 @@ public User getVipUser() {
 
     @Override
     public Collection<? extends GrantedAuthority> getAuthorities() {
-        return new ArrayList<>(); // not used at the moment
+        return List.of(new SimpleGrantedAuthority("ROLE_" + vipUser.getLevel().name().toUpperCase()));
```
axlbonnet: add a if `vipUser.getLevel() != null` before just in case.

vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/dao/mysql/EngineData.java
```diff
@@ -104,6 +112,23 @@ public List<Engine> get() throws DAOException {
         }
     }
 
+    @Override
+    public Engine get(String name) throws DAOException {
+        String query = "SELECT name, endpoint, status FROM VIPEngines WHERE name = ? ORDER BY name";
```
axlbonnet: You can remove `ORDER BY name`

vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/business/EngineBusiness.java
axlbonnet: I get it, no need to change anything yet as `getAll` is never called in a non-admin way.
However:
- i think `permissions.filter(PermissionChain::admin);` misses in `getAll` and `get`
- even though the controller and the security config protect these method in the REST calls, it is good to know which method are protected for internal calls. For instance `getUsableEngines` is not protected and it could be necessary to know for a future developer. So I suggest adding `VIPExternalSafe` on all the protected method, but i'm open to discuss that !

axlbonnet (commented):
Backend review

vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/security/common/SpringPrincipalUser.java
bergamote156: Ok, adapting SessionAuthenticationToken to use the user UserDetails role instead of the role passed.

vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/security/common/SpringPrincipalUser.java
```diff
@@ -26,7 +28,11 @@ public User getVipUser() {
 
     @Override
     public Collection<? extends GrantedAuthority> getAuthorities() {
-        return new ArrayList<>(); // not used at the moment
+        if (vipUser.getLevel() != null) {
+            return List.of(new SimpleGrantedAuthority("ROLE_" + vipUser.getLevel().name().toUpperCase()));
+        }
+        // level should not be null
```
axlbonnet: you can print a warning if you want.

vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/business/EngineBusiness.java
axlbonnet: Do you want to discuss that ?

axlbonnet (changes requested):
The authorities stuff for SessionAuthentication suits me.
Can you update `ApikeyAuthenticationToken` in the same way ?
Also, still a remark in EngineBusiness to handle.

vip-application/src/main/java/fr/insalyon/creatis/vip/application/server/business/EngineBusiness.java
bergamote156: I added it to stay coherent with the rest.

axlbonnet:
Good for the server part.
I still need to check and review the front part.

PR #698, opened 2026-08-20
fix sql injection

MayssaRouissi:


vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/dao/mysql/UsersGroupsData.java
```diff
@@ -196,31 +196,35 @@ public List<String> getUsersFromGroups(List<String> groups) throws DAOException
...
+        ps.setString(1, email);
 
-            List<Boolean> properties = new ArrayList<Boolean>();
+        try (ResultSet rs = ps.executeQuery()) {
             boolean isPublic = false;
 
             while (rs.next()) {
                 if (rs.getInt("public") == 1) {
                     isPublic = true;
+                    break;
```
axlbonnet: was it needed to change this method ?
i think you changed the behavior, but actually this method is not useful.
I suggest reverting your changes and we'll clean this method when we'll get rid of GWT.

vip-core/src/main/java/fr/insalyon/creatis/vip/core/server/dao/mysql/UsersGroupsData.java
```diff
@@ -160,32 +160,32 @@ public void setUserGroups(String email, Map<String, GROUP_ROLE> groups)
...
 
-            ResultSet rs = ps.executeQuery();
-            List<String> users = new ArrayList<String>();
-
-            while (rs.next()) {
-                users.add(rs.getString("first_name") + " "
-                        + rs.getString("last_name"));
+        try (PreparedStatement ps = getConnection().prepareStatement(query)) {
+            for (int i = 0; i < groups.size(); i++) {
+                ps.setString(i + 1, groups.get(i));
```
axlbonnet: You can do as in `ApplicationData::getApplicationsByGroups` for a more clarity.

axlbonnet (commented):
Seems fine, I suggest an minor improvement.

axlbonnet:
Nice ! Thanks !

